PR #8, opened 2014-07-09
Add an option in setup to use ProgramDataFolder for Vista and above

Goof-fr:
Add an option to use ProgramData Folder to store configuration files and logs
It will help with Microsoft Vista and above compatibility

### DIFF
--- a/windows-nsis/openvpn.nsi
+++ b/windows-nsis/openvpn.nsi
@@ -258,9 +258,14 @@ Section /o "${PACKAGE_NAME} Service" SecService
 	SetOutPath "$INSTDIR\bin"
 	File "${OPENVPN_ROOT}\bin\openvpnserv.exe"
 
-	SetOutPath "$INSTDIR\config"
-
-	FileOpen $R0 "$INSTDIR\config\README.txt" w
+	${If} ${SectionIsSelected} ${SecProgData}
+		SetOutPath "$APPDATA\${PACKAGE_NAME}\config"
+	${else}
+		SetOutPath "$INSTDIR\config"
+	${EndIf}
+	
+	CreateDirectory "$OUTDIR"
+	FileOpen $R0 "$OUTDIR\README.txt" w
 	FileWrite $R0 "This directory should contain ${PACKAGE_NAME} configuration files$\r$\n"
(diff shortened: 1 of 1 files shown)


mattock:
Sorry for the delay, I didn't have notifications turned on so I missed this and many other pull requests. In general this looks fine. However, before merging this I'd need confirmation of what happens if

1) The HKLM\Software\OpenVPN registry key is removed entirely (e.g. using regexit)
2) OpenVPN is installed to PROGRAMDATA using this new option
3) OpenVPN-GUI is launched as an admin

Does the first OpenVPN-GUI launch overwrite any of the registry keys your modified NSIS installer created? I'm not sure how OpenVPN-GUI behaves if some of its registry keys are present and some are missing.

mattock:
I had quick look at registry.c in openvpn-gui and it does not seem to overwrite existing registry keys. I will still have a look at what's happening on a real system.

Goof-fr:
This patch, does :
 - Create the directories into "ProgramData" folder for the "config" and "logs" folders
 - Create at installation the values in the registry that are loaded by OpenVPN and OpenVPN GUI to uses these directories instead of those created in the "Program files" folder.

The registry entries in "HKEY_LOCAL_MACHINE\SOFTWARE\OpenVPN" are mostly used by the OpenVPN Service when he had to look for Services .OVPN files to start automatically and for logs.

The registry entries in "HKEY_LOCAL_MACHINE\SOFTWARE\OpenVPN-GUI" are used by OpenVPN GUI to find the .OVPN configuration files and locate log files

mattock:
I built experimental installers based on your latest patches:

* http://build.openvpn.net/downloads/temp/openvpn-install-2.3.8-I148-i686.exe
* http://build.openvpn.net/downloads/temp/openvpn-install-2.3.8-I148-x86_64.exe

I tested the latter installer on a Windows 8.1 64-bit VM and immediately ran into issues. It does not seem to matter whether if SecProgData is checked or not: the registry keys will always end up pointing to the ProgramData folder. Can you reproduce this behavior yourself?

Besides this the code itself looks good. That said, would it be possible to create a prefix variable based on checked/unchecked value of the SecProgData, and use that to build the correct paths? This would make code slightly cleaner.

mattock:
Any comments on the issues I encountered? If not, I'll close this pull request.

PR #10, opened 2014-12-02
Fix the 'libpam required but missing' issue when cross-compile for ARM

yaoyaozong:
The is a problem when using the openvpn-build to cross-compile for arm platform on Ubuntu 14.
after fired the command
IMAGEROOT=`pwd`/image-arm CHOST=arm-linux-gnueabi CBUILD=x86_64-pc-linux-gnu ./build

the error 'libpam required but missing' will shows up.
This is because there is NO libpam library in the cross-compile environment.

Considering the libpam is not common used(I suppose). I disabled it in the default options.

Cheers,
Billy YAO


mattock:
We [had a discussion](http://thread.gmane.org/gmane.network.openvpn.devel/10984) about this pull request on the mailing list. The end result was that we don't want to disable PAM by default, because it might break things elsewhere. However, this problem [has now been documented](https://community.openvpn.net/openvpn/wiki/BuildingUsingGenericBuildsystem) in openvpn-build documentation.

PR #17, opened 2016-01-11
travis-ci config

chipitsine:
travis-ci allows automation of pull request build testing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,5 +24,6 @@ before_script:
     - unset CC; unset CXX
 
 script:
+    - ls
     - cd openvpn-build/generic
     - ./build
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,4 @@ before_script:
     - unset CC; unset CXX
 
 script:
-    - ls
-    - cd openvpn-build/generic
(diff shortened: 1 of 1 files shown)


mattock:
This looks like a good starting point. Later we could go with build-complete, which would catch even more issues. Based on Travis-CI  that should be doable.

chipitsine:
what is "build-complete" ?

mattock:
See the [build documentation](https://community.openvpn.net/openvpn/wiki/BuildingUsingGenericBuildsystem#CreatingaNSISinstallerwindows-nsissubdir). That said, converting build-complete build process to Travis-CI would be way more difficult, because we'd need to install a bunch of dependencies, run custom scripts, etc. Plus my Windows "buidslave" already runs build-complete/build-snapshot against "master" and "release/2.3" branches on new commits, so I think we're fine.

chipitsine:
if you mean creating nsis installer on travis, I think it would be overkill and not very neccessary.
currently, I think, travis will build pull request and tell us whether build is passing or not. let us stick on that thing first.

mattock:
Yes, exactly. We're on the same page then.

PR #18, opened 2016-02-17
Fail if trying to run I60x installer on Windows XP

mattock:
Trac #632

Tested-by: Fish Wang <fish.thss@gmail.com>
Tested-by: Jan Just Keijser <janjust@nikhef.nl>

### DIFF
--- a/windows-nsis/openvpn.nsi
+++ b/windows-nsis/openvpn.nsi
@@ -28,6 +28,11 @@ SetCompressor lzma
 !addplugindir .
 !include "nsProcess.nsh"
 
+; WinVer.nsh is needed for detecting the Windows version and bailing out if the
+; installation cannot possibly succeed, e.g. when using I60x installers that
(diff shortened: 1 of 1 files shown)


mattock:
Here's the code to fix https://community.openvpn.net/openvpn/ticket/632. I made this a PR because I think it's a good practice; we should have some form of review for all commits. Plus this way we get Travis build tests for free (not that it helps in this particular case).

@chipitsine: care to take a quick look?

mattock:
I asked a guy who wrote some useful NSI stuff in the past to review the new NSI code. If that plan falls through, I'll apply a "lazy-ACK" and merge this, as the new functionality has been tested independently by three different people.

mattock:
I'll apply lazy-ACK and merge this. If there are issues with this commit, the [Git snapshot builds](http://build.openvpn.net/downloads/snapshots/) should reveal them before hitting the OpenVPN 2.4 alphas.

PR #21, opened 2016-06-20
fix build in case no wget is installed (using curl)

chipitsine:
```
R00-2+jenkins@r00-2 ~/openvpn-build/generic
$  IMAGEROOT=`pwd`/image-win32 CHOST=i686-w64-mingw32 CBUILD=i686-pc-cygwin ./build
./build: line 77: cd: /home/jenkins/openvpn-build/generic/sources: No such file or directory
FATAL: Cannot download http://www.oberhumer.com/opensource/lzo/download/lzo-2.09.tar.gz

R00-2+jenkins@r00-2 ~/openvpn-build/generic
$
```

### DIFF
--- a/generic/build
+++ b/generic/build
@@ -365,8 +365,8 @@ solaris_common() {
 main() {
 	export CFLAGS="${TARGET_CFLAGS} ${EXTRA_TARGET_CFLAGS}"
 	export LDFLAGS="${TARGET_LDFLAGS} ${EXTRA_TARGET_LDFLAGS}"
-	download
 	create_layout
+	download
 	extract
 	build_dep
 	if [ -z "${DO_ONLY_DEPS}" ]; then
--- a/generic/build.vars
+++ b/generic/build.vars
@@ -8,7 +8,7 @@ TAP_WINDOWS_VERSION="${TAP_WINDOWS_VERSION:-9.21.2}"
 OPENVPN_VERSION="${OPENVPN_VERSION:-2.3.11}"
 OPENVPN_GUI_VERSION="${OPENVPN_GUI_VERSION:-10}"
 
-OPENSSL_URL="${OPENSSL_URL:-http://www.openssl.org/source/openssl-${OPENSSL_VERSION}.tar.gz}"
+OPENSSL_URL="${OPENSSL_URL:-https://www.openssl.org/source/openssl-${OPENSSL_VERSION}.tar.gz}"
 PKCS11_HELPER_URL="${PKCS11_HELPER_URL:-http://downloads.sourceforge.net/project/opensc/pkcs11-helper/pkcs11-helper-${PKCS11_HELPER_VERSION}.tar.bz2}"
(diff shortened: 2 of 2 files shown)


generic/build
```diff
@@ -74,7 +74,10 @@ geturl() {
 		"${WGET}" ${WGET_OPTS} --directory-prefix="${SOURCESROOT}" "${url}"
 	else
 		(
-			cd "${SOURCESROOT}" && \
+		if [ ! -d "${SOURCESROOT}" ]; then
+			mkdir -p "${SOURCESROOT}"
```
mattock: Interesting...looking at _generic/build_ I see that the create_layout() function creates SOURCESROOT, but it runs _after_ the download() function. Have you tried reverting the position of those functions, so that the main() function would look like this?
```
main() {
        export CFLAGS="${TARGET_CFLAGS} ${EXTRA_TARGET_CFLAGS}"
        export LDFLAGS="${TARGET_LDFLAGS} ${EXTRA_TARGET_LDFLAGS}"
        create_layout
        download
        extract
        build_dep
        --- snip ---
```
Would that solve this problem?

generic/build.vars
```diff
@@ -8,7 +8,7 @@ TAP_WINDOWS_VERSION="${TAP_WINDOWS_VERSION:-9.21.2}"
 OPENVPN_VERSION="${OPENVPN_VERSION:-2.3.11}"
 OPENVPN_GUI_VERSION="${OPENVPN_GUI_VERSION:-10}"
 
-OPENSSL_URL="${OPENSSL_URL:-http://www.openssl.org/source/openssl-${OPENSSL_VERSION}.tar.gz}"
+OPENSSL_URL="${OPENSSL_URL:-https://www.openssl.org/source/openssl-${OPENSSL_VERSION}.tar.gz}"
```
mattock: This looks reasonable, but I suggest renaming the commit message and making it fit into about 72 characters. Right now it overflows and will make thinks like "git shortlog" look silly.

Maybe just change the commit message to "Fetch OpenSSL sources via https"?

mattock:
Commit https://github.com/OpenVPN/openvpn-build/pull/21/commits/c1263a9b67c7860e4db0abfa508575644fd950ce seems reasonable, but the commit message should be improved. I suggest this: "Ensure that curl follows redirects".

generic/build
chipitsine: yes, it also helps.

chipitsine:
thoughts ?

generic/build
```diff
@@ -365,8 +365,8 @@ solaris_common() {
 main() {
 	export CFLAGS="${TARGET_CFLAGS} ${EXTRA_TARGET_CFLAGS}"
 	export LDFLAGS="${TARGET_LDFLAGS} ${EXTRA_TARGET_LDFLAGS}"
-	download
 	create_layout
+	download
 	extract
```
mattock: I would keep this part in a separate commit, because it's entirely unrelated to the curl stuff. Maybe in a commit called "Create layout before downloading dependency packages"?

mattock:
The contents of this PR look good, so once the download/create_layout part is split into a separate commit (see my line note) I'll merge this. Later we can figure out whether we need both curl _and_ wget support.

chipitsine:
any news on that ?

PR #23, opened 2016-06-30
Register openvpnserv2 as a service during install

mattock:
This PR makes the install script to register openvpnserv2 as a Windows system service. For this change to make sense the code for the old openvpnserv.exe needs to be modified slightly - that will be fixed in another patch. Right now openvpnserv2 sets its StartupType to "automatic", which is not what we want, and which needs to be fixed. That fix is, however, unrelated to this PR.

### DIFF
--- a/windows-nsis/openvpn.nsi
+++ b/windows-nsis/openvpn.nsi
@@ -71,9 +71,6 @@ InstallDirRegKey HKLM "SOFTWARE\${PACKAGE_NAME}" ""
 
 !define MUI_COMPONENTSPAGE_SMALLDESC
 !define MUI_FINISHPAGE_SHOWREADME "$INSTDIR\doc\INSTALL-win32.txt"
-!define MUI_FINISHPAGE_RUN_TEXT "Start OpenVPN GUI"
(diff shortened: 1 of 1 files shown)


mattock:
Added better Interactive Service integration as a separate patch, as agreed upon in [here](http://thread.gmane.org/gmane.network.openvpn.devel/12010).

mattock:
Fixed [the problems](http://thread.gmane.org/gmane.network.openvpn.devel/12010) Selva identified, namely
* sc.exe would fail if SecService was not selected, but SecInteractiveService was
* Installer would restart OpenVPN-GUI as admin, which, with Interactive Service in place, no longer makes sense.

mattock:
An installer with all the above fixes is available here:
* http://build.openvpn.net/downloads/snapshots/openvpn-install-2.3_git-I907-x86_64.exe

The installer was announced on openvpn-devel mailing list at the time of my previous comment. More testing would be good to have, so that I don't have to merge PR this with lazy-ACK.

mattock:
Merging with lazy-ACK.

PR #24, opened 2016-08-23
some cleanup (removed snappy compression) + correct openvpn version

chipitsine:


windows-nsis/build-snapshot
```diff
@@ -31,7 +32,7 @@ main() {
 	git clone -q --depth 1 --branch "${OPENVPN_BRANCH}" "${OPENVPN_URL}" || die "openvpn.clone"
 	cd openvpn
 	autoreconf -ivf || die "autoreconf.openvpn"
-	./configure --disable-lzo --disable-snappy --disable-lz4 || die "openvpn.configure"
+	./configure --disable-lzo --disable-lz4 || die "openvpn.configure"
```
mattock: This makes sense in the "master" branch indeed.

Just so that you know: yesterday I created a "release/2.3" branch in openvpn-build, because the most recent changes to openvpn.nsi were incompatible with OpenVPN 2.3.x. The HEAD of the release/2.3 branch is now at 42aed3b6e1267258 ("Revert "Fail if trying to run I60x installer on Windows XP""), but some of the more recent commits could probably be cherry-picked and applied to release/2.3 too. Feel free to experiment.

chipitsine: it's "build-snapshot" script, it is not assumed for release/2.3 ?

windows-nsis/build-snapshot
```diff
@@ -12,6 +12,7 @@ get_full_path() {
 }
 
 GIT="${GIT:-git}"
+OPENVPN_VERSION="${OPENVPN_VERSION:-2.3_git}"; export OPENVPN_VERSION
```
mattock: Having a default for OPENVPN_VERSION makes sense imho. Why do you export the variable, though?

windows-nsis/build-snapshot
mattock: The build-snapshot script is primarily intended for Git "master", although it could be used with release branches. In any case this change is good.

Based on Git logs snappy support was never added to the OpenVPN's "release/2.3" branch (commit 38d96bd7975 in "master"), nor removed from it (commit 9403e3f4b510fb in "master"). I also verified that build-complete does not break with this change. Feel free to create another PR for removing --disable-snappy from openvpn-build's "release/2.3" branch.

windows-nsis/build-snapshot
chipitsine: it is bash variable otherwise, it does not work. we need environment variable here

windows-nsis/build-snapshot
chipitsine: ok, I will add another PR for release/2.3 branch, probably. or you can do it yourself.

PR #25, opened 2016-08-23
travis-ci: refactoring

chipitsine:


mattock:
The idea and content of this PR looks good, but the title is somewhat confusing. Afaics, the main purpose of this PR/commit is not to refactor travis.yml, but to add support for testing build-snapshot and build-complete with Travis. Could you rename the commit so that it is clear from the commit logs what is being done?

I noticed that [build-snapshot fails](https://travis-ci.org/OpenVPN/openvpn-build/jobs/154576583#L5315) because openvpn-2.3.11 directory does not exist. Was this the reason you made [this change](https://github.com/OpenVPN/openvpn-build/pull/24/files#diff-bf5c527fb371957e030ec79acf34ab3dR15)?

chipitsine:
yes

mattock:
Just for the record: the Travis-CI build fails, but the reason for the failure has been fixed in another PR.

PR #27, opened 2016-08-29
Switch to openvpnserv2-1.1.0.0.exe

mattock:
The previous openvpnserv2 version tried to start OpenVPNService immediately
after registering itself as a Windows service. This failed, as tap-windows6 was
not properly setup yet. This produced quite confusing errors messages when
OpenVPN was being installed.


mattock:
The openvpnserv2 project now [has the fix](https://github.com/xkjyeah/openvpnserv2/pull/8), and the fixed version is 1.1.0.0. I will keep this PR open for a few days and then merge it.

selvanair:
Independent of the fix in openvpnserv2, can't we move the service install (in the nsis script) to after tap install to avoid such failures?

mattock:
@selvanair: yes, that would have been one option, but openvpnserv2's misbehavior had to be fixed anyways. Now the "-install" flag behaves the same as in openvpnserv.exe, i.e.the service is installed, but not started.

selvanair:
Aha, I misunderstood. Sure, we don't want the service to start automatically on -install.

mattock:
Now that this is sorted out I'll merge this PR.

PR #29, opened 2016-09-16
travis-ci: we need wait for 16.04 images

chipitsine:
mingw 4.8 is not usable for compiling Vista things like firewall
I commented snapshot build for a while (I was not lucky in installing 4.9 on ubuntu 14.04)

let us wait for 16.04 images


chipitsine:
I've managed to be lucky with "xenial", not "wily"

https://travis-ci.org/chipitsine/openvpn-build/jobs/160679293

however, I picked openvpn-master from http://build.openvpn.net with (wrong) c99 patch.

let us wait until master tarball gets updated :-)

PR #32, opened 2016-09-27
Use more reasonable build parameters in build.vars

mattock:
Previously the installers generated using "master" branch in openvpn-build would
not work properly. With this change all the components (openvpn, openvpnserv,
openvpn-gui) work in unison. In addition build should work out of the box
without any modifications of build.vars or build-complete.vars.

### DIFF
--- a/generic/build.vars
+++ b/generic/build.vars
@@ -1,18 +1,18 @@
 # -*- mode: sh; -*-
 BUILD_VERSION="001"
 
-OPENSSL_VERSION="${OPENSSL_VERSION:-1.0.1t}"
+OPENSSL_VERSION="${OPENSSL_VERSION:-1.0.1u}"
 PKCS11_HELPER_VERSION="${PKCS11_HELPER_VERSION:-1.11}"
 LZO_VERSION="${LZO_VERSION:-2.09}"
 TAP_WINDOWS_VERSION="${TAP_WINDOWS_VERSION:-9.21.2}"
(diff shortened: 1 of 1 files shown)


mattock:
One more thing I should probably mention... openvpn-2.3_git.tar.gz will now always contain the latest Git "master" version of OpenVPN, with a delay of <1 hour. Update of the tarball is handled on the webserver using a cronjob. This also explains why the URL is now pointing to /snapshots/openvpn-2.3_git.tar.gz instead of /releases/openvpn-2.3_git.tar.gz.

mattock:
Regarding the build failures... is there a problem with mingw-w64 headers?

    In file included from /usr/share/mingw-w64/include/ipsectypes.h:8:0,
                     from /usr/share/mingw-w64/include/fwpmtypes.h:9,
                     from /usr/share/mingw-w64/include/fwpmu.h:10,
                     from block_dns.c:40:

    /usr/share/mingw-w64/include/iketypes.h:747:2: error: #endif without #if
     #endif /* __iketypes_h__ */

chipitsine:
sorry for that.
somehow I used to think I did that PR already

https://github.com/OpenVPN/openvpn-build/pull/33

there's Vista firewall manipulation, which is not supported on mingw gcc 4.8
I tried adding xenial repo, it is not very stable

I think, the best for now would be commenting it for a while.

I will fix it later

mattock:
I'll merge this PR as it's unrelated to the Travis build failure.

PR #34, opened 2016-10-13
Make output of "sc.exe start OpenVPNServiceInteractive" less verbose

mattock:
If the command succeeds, it outputs lots of diagnostic stuff, as if it had
errored out. This change takes the return code of the command and produces a
short, informational message based on it.


mattock:
64-bit installer that contains this functionality is available [here](http://build.openvpn.net/downloads/snapshots/openvpn-install-2.3_git-I605-do-ifconfig-after-tun-v2-sc-exe-x86_64.exe). I tested this pretty well, but only on Windows 7. So some independent testing would be nice to have.

mattock:
Merging with lazy-ACK.

PR #35, opened 2016-10-19
Migrate INSTALL-win32.txt from the OpenVPN project and modernize it.

mattock:
Migrate INSTALL-win32.txt from the OpenVPN project and modernize it.

### DIFF
--- a/windows-nsis/INSTALL-win32.txt
+++ b/windows-nsis/INSTALL-win32.txt
@@ -1,0 +1,70 @@
+GENERAL QUICKSTART FOR WINDOWS
+
+OpenVPN requires a configuration file and key/certificate files. You should 
+obtain these and save them to OpenVPN's configuration directory, usually 
+C:\Program Files\OpenVPN\config. Starting with OpenVPN 2.4-alpha1 you can also 
+put user-specific configurations to %USERPROFILE%\OpenVPN\config when using 
+OpenVPN GUI.
+
+You can run OpenVPN as a Windows system service, by using OpenVPN GUI or from a 
+command-prompt. To use the OpenVPN GUI, double click on the desktop icon or 
+start menu icon. OpenVPN GUI is a system-tray applet, so an icon for the GUI 
+will appear in the lower-right corner of the screen. Right click on the system 
+tray icon, and a menu should appear showing the names of your OpenVPN 
+configuration files, and giving you the option to connect. Starting in OpenVPN 
+2.4-alpha1 admin privileges are no longer required for launching OpenVPN 
+connections using configuration files stored in the global config directory 
+(usually C:\Program Files\OpenVPN\config). Users who belong to the built-in 
+administrator group, or to the local "OpenVPN Administrator" group can also 
+store configuration files under %USERPROFILE%\OpenVPN\config. In both of these 
+cases OpenVPN Interactive Service needs to be running: if it is not, 
+administrator privileges are still required to successfully start OpenVPN 
+connections. For further details please refer to OpenVPN GUI documentation:
+
+    https://github.com/OpenVPN/openvpn-gui
+
+OpenVPN 2.4-alpha1 and later come bundled with three system services:
+
+    OpenVPNService (new, improved background service)
+    OpenVPNServiceInteractive (required to run GUI as non-admin)
+    OpenVPNServiceLegacy (deprecated background service)
+
+The installer starts OpenVPNServiceInteractive automatically and configures it 
+to start at system startup. This is done to allow unprivileged users to start 
+OpenVPN connections using OpenVPN GUI without any extra configuration.
+
+OpenVPNService is based on openvpnserv2, a complete rewrite of the OpenVPN 
+service wrapper. It is intended for launching OpenVPN instances that should be 
+up at all times, instead of being manually launched by a user. OpenVPNService is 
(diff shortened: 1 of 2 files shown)


windows-nsis/INSTALL-win32.txt
```diff
@@ -0,0 +1,62 @@
...
+OpenVPN GUI.
+
+You can run OpenVPN as a Windows system service, by using OpenVPN GUI or from a 
+command-prompt. To use the OpenVPN GUI, double click on the desktop icon or 
+start menu icon. OpenVPN GUI is a system-tray applet, so an icon for the GUI 
+will appear in the lower-right corner of the screen. Right click on the system 
+tray icon, and a menu should appear showing the names of your OpenVPN 
+configuration files, and giving you the option to connect. Starting in OpenVPN 
+2.4-alpha1 it is no longer necessary to have admin privileges to run the GUI, as 
+long as the OpenVPN Interactive Service is running. For details please refer to
```
schwabe: Reference that this group can be modified and who is in there by default?

windows-nsis/INSTALL-win32.txt
```diff
@@ -0,0 +1,62 @@
+GENERAL QUICKSTART FOR WINDOWS
+
+OpenVPN requires a configuration file and key/certificate files. You should 
+obtain these and save them to OpenVPN's configuration directory, usually 
+C:\Program Files\OpenVPN\config. Starting with OpenVPN 2.4-alpha1 you can also 
+put user-specific configurations to %USERPROFILE%\OpenVPN\config when using 
+OpenVPN GUI.
```
schwabe: why are cluttering the user's home and putting that into %APPDATA%\OpenVPN\config? Or have that at least as alternative?

windows-nsis/INSTALL-win32.txt
```diff
@@ -0,0 +1,62 @@
...
+OpenVPN GUI's documentation:
+
+    https://github.com/OpenVPN/openvpn-gui
+
+OpenVPN 2.4-alpha1 and later come bundled with three system services:
+
+    OpenVPNService (new, improved background service)
+    OpenVPNServiceInteractive (to run GUI as non-admin)
+    OpenVPNServiceLegacy (deprecated background service)
+
```
schwabe: Add a note here like "Background service is used for OpenVPN connection that should be always running or OpenVPN services processes" to give the user an idea what background service is good for?

windows-nsis/INSTALL-win32.txt
```diff
@@ -0,0 +1,62 @@
...
+The installer starts OpenVPNServiceInteractive automatically and configures it 
+to start at system startup. This is done to allow unprivileged users to start 
+OpenVPN connections using OpenVPN GUI without any extra configuration.
+
+The new OpenVPNService is a complete rewrite, and is able restart crashed 
+OpenVPN connections. It also works properly on new Windows versions unlike 
+OpenVPNServiceLegacy, which is included mostly for historic reasons.
+
```
schwabe: mostly? So there are there other reason that you don't tell me?

mattock: Well, I'm not sure that we even should include OpenVPNServiceLegacy, or that "historic" reasons is a valid reason. However, the decision was made to include it, so in it goes. The theory probably was that if openvpnserv2 somehow manages to behave worse than old openvpnserv's "automatic" part, then user can fall back to OpenVPNServiceLegacy. I'll modify the text to make that clear.

windows-nsis/INSTALL-win32.txt
mattock: Good point. Will add.

windows-nsis/INSTALL-win32.txt
schwabe: I am not so familiar with the code. Could we enable both directories?

mattock: Scanning both seems quite easy, based on a cursory look at [openvpn_config.c](https://github.com/OpenVPN/openvpn-gui/blob/master/openvpn_config.c#L220). The question then becomes: do we _want_ to have two places where users can put user-specific config files? If we remove either one later we will break user configs.

mattock: In the "Settings -> Advanced" tab of the latest OpenVPN-GUI 11 the user can select where configuration and log files are stored. Still, having a sane default is important.

mattock:
@schwabe : I clarified the parts you commented about. Does it look good now?

windows-nsis/INSTALL-win32.txt
mattock: I sent email to openvpn-devel about this.

mattock: The end result of the [above email discussion](https://www.mail-archive.com/openvpn-devel@lists.sourceforge.net/msg12738.html) was that %USERPROFILE%\OpenVPN\config is probably the least bad solution for now. The "correct" place for the config files would be under %USERPROFILE%\Appdata, but that directory is hidden on the filesystem level, so users would not see it, even though it _is_ accessible, e.g. using _cd $HOME\Appdata_ in Powershell. The idea behind hiding the directory probably was that applications, not users, would directly modify that directory. So we'll go with the original plan.

PR #36, opened 2016-10-19
Update build parameters to match openvpn-2.4_alpha2

mattock:
Currently OpenVPN version is set to 2.4_alpha2, even in Git "master". Make OPENVPN_VERSION in build.vars consistent with OpenVPN's versioning.


mattock:
Note that openvpn tarballs are still loaded from the "snapshots" directory on build.openvpn.net. The rationale is that the "master" branch in openvpn-build will build the latest Git version of OpenVPN that has version set to 2.4_alpha2. So any commits added prior to the next alpha/beta will also get in to installers.

mattock:
Merging with lazy-ACK.

PR #43, opened 2016-11-03
Fix linefeeds in INSTALL-win32.txt

mattock:
The INSTALL-win32.txt that was migrated from openvpn had UNIX linefeeds. As this
file is supposed to be shown to users using notepad.exe, it needs Windows
linefeeds. This patch fixes the "master" branch, which had a heavily modified
version of INSTALL-win32.txt.


selvanair:
Desirable whitespace change (unix2dos). ACK.

PR #47, opened 2016-11-17
Update OpenVPN version to 2.4_beta1

mattock:
This is necessary to ensure that build-snapshot continues to work, and to ensure that people build the latest version by default.


mattock:
Nobody has complained, plus these changes are trivial and necessary. Merging.

PR #48, opened 2016-11-19
check for Windows Vista

chipitsine:


chipitsine:
actually, we can provide the proper http link for windows xp installer
@mattock ?

it is "discussion" PR. I consider "sorry, it does not work" as bad UX.
it should be like "ok, you seem to be running not supported OS, here's link to old installer"

windows-nsis/openvpn.nsi
```diff
@@ -428,6 +428,11 @@ Function .onInit
 	${GetParameters} $R0
 	ClearErrors
 
+${IfNot} ${AtLeastWinVista}
+	MessageBox MB_OK "Windows Vista and above required"
+	Quit
+${EndIf}
+
```
mattock: Spaces instead of tabs (or vice versa) here? Otherwise looks like it should work (have not tested yet).

chipitsine: ok, I'll change to tabs

we might want to add some CI (??), which will check for tabs/spaces/etc

mattock:
I agree that a link to Windows XP installer would make sense. A few options:
* Link to the [download page](https://openvpn.net/index.php/download/community-downloads.html)
* Add direct links (32/64-bit) to "latest Windows XP installer" to [this webserver directory](http://build.openvpn.net/downloads/releases/latest/) and link to those in the installer

I'd prefer the latter approach, but I'm fine with both.

Does the message box element support HTML links, or will they be shown as plain text?

chipitsine:
will "the latest stable"  always work on WinXP ?

chipitsine:
or, there will be "the latest WinXP installer" ?

mattock:
The latest stable link will not work for Windows XP. The current link is pointing to OpenVPN 2.3.x installer, but one with tap-windows6 driver. I will add special links that are Windows XP-specific (2.3.x + tap-windows).

mattock:
There are now static links to the latest Windows XP installers:
* http://build.openvpn.net/downloads/releases/latest/openvpn-install-latest-winxp-x86_64.exe
* http://build.openvpn.net/downloads/releases/latest/openvpn-install-latest-winxp-i686.exe

PR #50, opened 2016-11-21
added "launch on windows startup" to installer

chipitsine:


windows-nsis/openvpn.nsi
```diff
@@ -338,6 +340,11 @@ Section /o "${PACKAGE_NAME} GUI" SecOpenVPNGUI
 		CreateShortCut "$SMPROGRAMS\${PACKAGE_NAME}\${PACKAGE_NAME} GUI.lnk" "$INSTDIR\bin\openvpn-gui.exe" ""
 		CreateShortcut "$DESKTOP\${PACKAGE_NAME} GUI.lnk" "$INSTDIR\bin\openvpn-gui.exe"
 	${EndIf}
+
+	${If} ${SectionIsSelected} ${SecLaunchGUIOnStartup}
+		WriteRegStr HKCU "Software\Microsoft\Windows\CurrentVersion\Run" "OpenVPN-GUI" "$INSTDIR\bin\openvpn-gui.exe"
```
selvanair: This sets to start on logon for the user with whose rights the installer is being run. In a sane Windows setup the user is a limited user and has to input an admin username/password to run the installer.  Then the exec is added to startup for the wrong user. But I know most users are admins on Windows and then this just works.

PR #52, opened 2016-11-23
openvpn-2.4 requires at least Windows Vista, added check to installer

chipitsine:


mattock:
Excellent, thanks! I will produce a test installer with this PR and ask people on openvpn-devel to test it. Hopefully we can merge this early next week, before the next 2.4.x release.

windows-nsis/openvpn.nsi
```diff
@@ -454,6 +454,26 @@ Function .onInit
...
+	Goto guiEndYes
+
+	guiEndNo:
+	Quit
+
+	guiEndYes:
+	${If} ${RunningX64}
+		ExecShell "open" "http://build.openvpn.net/downloads/releases/latest/openvpn-install-latest-winxp-x86_64.exe" 
+	${Else}
+		ExecShell "open" "http://build.openvpn.net/downloads/releases/latest/openvpn-install-latest-winxp-x86_64.exe"
```
chipitsine: sorry for copy/paste

as installer itself is signed, I consider http as acceptable, @mattock ?

mattock:
@chipitsine : for the final version please change the links from http to https. I will include this version in the test installers I'm building today. The installers will also have openvpn-gui.exe patched with most of the currently open PRs.

mattock:
@chipitsine : I did not notice your comment about http. Anyways, I think https is still better, even if the installers are signed.

chipitsine:
I think, it does not matter.

1) (bad points) somewhere you cannot download from https due to TLS misconfiguration, but it is very rare
2) (good point) marketing people like https, I've no idea why. people do not submit any personal or sensitive information (credit card details, etc), installer is signed.

we can switch to https if you want, I do not care

mattock:
Yes, from a technical viewpoint you're entirely correct. However, some people will complain about the http links either because they're worried, or because they think using http is a bad practice. Then we have to spend time explaining why http is not necessary in this case.

This is pretty much the same situation as with OpenSSL releases: if I don't release new OpenVPN installer with the updated OpenSSL version, people will always start asking "is OpenVPN affected?", even if none of the vulnerabilities actually affect OpenVPN. Responding to those questions takes more time in the long run than just releasing new installers.

So let's just give up now and make the links https :).

mattock:
We can merge this [when somebody reports](https://www.mail-archive.com/openvpn-devel@lists.sourceforge.net/msg13256.html) that this fails (as expected) on Windows XP

mattock:
The installer fails as expected on Windows XP, and provides the link to WinXP installers, and the link works. Tested by [debbie10t](https://www.mail-archive.com/openvpn-devel@lists.sourceforge.net/msg13258.html) and [Gert](https://www.mail-archive.com/openvpn-devel@lists.sourceforge.net/msg13267.html).

Merging.

PR #53, opened 2016-11-25
Update build parameters and build-snapshot to OpenVPN 2.4_beta2

mattock:


mattock:
To avoid having build-snapshot in a broken state for too long I'll merge this without a formal ACK. The changes should be trivial enough.

PR #54, opened 2016-11-25
check for .net 4.0 if openvpnserv2 is selected

chipitsine:


mattock:
The [NsisDotNetChecker](https://github.com/ReVolly/NsisDotNetChecker) seems to fairly active project, and the changes to openvpn.nsi are quite minor, so I'm generally fine with this. Some independent smoke testing is in order, though. I will add this PR to an installer for testing.

windows-nsis/openvpn.nsi
```diff
@@ -30,6 +30,10 @@ SetCompressor lzma
 !addplugindir .
 !include "nsProcess.nsh"
 
+; DotNetChecker.nsh to detect whether .net 4.0 is enabled, which is required for openvpnserv2 ( https://github.com/ReVolly/NsisDotNetChecker )
+!addplugindir .
```
chipitsine: It is added for safety. If we will ever remove nsProcess, this line must be kept

chipitsine:
I tested on win7 without .net 4.
If someone wants to test, there are plenty of free vm images at http://dev.modern.ie

windows-nsis/openvpn.nsi
mattock: Based on [NSIS documentation](http://nsis.sourceforge.net/Docs/Chapter5.html#addplugindir) !addplugindir makes NSIS scan a directory (here ".") for plugins. After the scan NSIS should see all of the plugins (here .nsh files) in the current directory. Unless you know something I don't, I'd say scrap the second !addplugindir command, and just keep the one above _!include "nsProcess.nsh"_.

chipitsine: do you mean "leave just single !addplugindir ." line instead of two ?

mattock: Yes, one should be enough.

mattock: Maybe something like this, then:
```
# !addplugindir ensures that nsProcess.nsh and DotNetChecker.nsh can be included
!addplugindir .
!include "nsProcess.nsh"
; DotNetChecker.nsh to detect whether .net 4.0 is enabled, which is required for openvpnserv2 ( https://github.com/ReVolly/NsisDotNetChecker )
!include "DotNetChecker.nsh"
```

chipitsine: ok

chipitsine:
are we ready to merge it?

mattock:
@chipitsine : uh, not yet, I didn't have any time to test this yet. I need to push out OpenVPN 2.4_rc1 out _now_, so let's get this to rc2, which will be released in ~2 weeks.

chipitsine:
I wish to get "stop (any) openvpn service if running before install" for rc2 :) too many things for rc2

selvanair:
We already remove the service in serction -pre, so just moving it up before checking for running openvpn.exe would be useful. And its a minor change.

That said the current approach of removing the service unconditionally even if the user selects not install the service does not look proper. What about change those lines in -pre to just stop the services and remove them only  in the service install section? When one selects not to install the service or the GUI they many not expect to have existing ones to be removed.

chipitsine:
what we do in a wrong way, we stop service asynchronously. we should wait for "stopped" state.

selvanair:
Where do we stop the service in the installer? We only remove it (which is synchronous) or start it (which is async, but that's OK). 

Anyway, my suggestion is to replace the 'openvpnservice -remove' in "Section -pre" with 'net stop openvpnservice /yes', not 'sc stop openvpnservice'.  The net command will wait for the service to stop. In this case we have to try stop 3 services: OpenVPNService, OpenVPNServiceInteractive and OpenVPNServiceLegacy and ignore any errors.

That plus stopping the GUI will likely kill all openvpn.exe processes. If any process is still running, we bail out as before with a message that openvpn is running etc..

selvanair:
Tested on windows 7 after removing .NET 4+ versions
It works but the user experience is not that great (not a fault of this PR):

The installation time increased from 30 sec to 6 minutes, not including the download time for dotnetfx.exe, extra 50MB download.

In my case, the first time I left it unattended and it rebooted after .NET install and the rest of the installation did not complete after the reboot. The installer was on a network drive, may be that explains the failure to contiune after restart..

Removed .NET4 and tried again with installer copied to local drive and this time it succeeded without a restart. In any case .NET updates could lead to restarts unless all existing programs using a previous version of .NET are not stopped. I would suggest not to install the runtime but to just pop up a  warning at the end if .NET 4.0 is not found, saying openvpnservice will not work until NET 4.0 (or later ?) is installed.

chipitsine:
missing .NET 4 is very rare case, so "any user experience" should be enough. even "not great user experience"

selvanair:
Restart without any warning or prompt? That is unacceptable. I tried again installing the dotnet exec from command line and it restarted the computer all of a sudden. Adding /promptrestart to the command line used in the macro may help? I have no appetite left for testing that.

After removing .NET4.5 that was installed on this machine many programs failed to start on reboot but I got a prompt saying something like .NET 4.5 required, do you want to install it? Making openvpnserv2 do something like that instead of a less helpful message that essentially says ".. contact the application developer ..."  would be saner approach.

chipitsine:
@selvanair , do you suggest to call DotNetChecker::IsDotNet${FrameworkVersion}Installed directly ?

I see no other option here: https://github.com/ReVolly/NsisDotNetChecker/blob/master/nsis/DotNetChecker.nsh

as for uninstalling .net **after** openvpserv2 is installed, do you also mean that ? I've not idea why does installer should take care of that. And what if someone will ever decide to remove openvpnserv2.exe manually ? There are a lot of possibilities to shoot yourself in the leg, just few of them are solved by installer

selvanair:
As /promptrestart I meant adding that to the command line used to install the downloaded net installer:
currently it is ExecWait '$TEMP\dotnetfx.exe /q /c:"install /q"' (line 100 of the nsh fiel you linked). Here /q is for quiet which I think suppresses any prompts for restart. I tried /? on that dotnetfx.exe file and found /promptrestart as an option. But no idea whether that really avoids restarts or not. Hard to test as restarts are not always triggered.

Ideally a .NET runtime installer should enumerate currently running programs that use .NET and offer to close them to avoid restarts. Again no idea how to integrate that with nsis.

> as for uninstalling .net after openvpserv2 is installed, do you also mean that ? I've not idea why does installer should take care of that. And what if someone will ever decide to remove openvpnserv2.exe manually ?

I did not mention any uninstallation. What I meant was openvpnserv2 could be programmed so as to pop up a message saying it needs net version xx to run and offer to install it if needed. Then we can completely avoid .NET runtime installation in our nsis installer.  I have seen net programs doing that and one usually gets directed to a link like http://dotnetsocial.cloudapp.net/GetDotnet?rClrVer=v4.0 To see how such a prompt will look like see https://blogs.msdn.microsoft.com/dotnet/p/targetingdotnet/
 
I'm not a .NET developer so no idea how to make openvpnserv2 do that. Currently if you try to start it without right version of .NET installed, it shows a stock message that says ".. contact the app developer.." or something like that.

chipitsine:
@selvanair  , I add app.config  https://github.com/chipitsine/openvpnserv2/commit/2ac3eabd3932344da1d7432cba60dd23ff653286

also, if you do not want to compile, you can download artifact here: https://ci.appveyor.com/project/chipitsine/openvpnserv2/build/artifacts

can you try it ?

selvanair:
@chipitsine Tested your openvpnserv2 build after removing .NET completely and with .Net 3.5 installed on Windows7. In all cases I get the following message as before:
net4-needed

For some .NET programs needing 4.0 I could earlier get a window like the one shown here  https://blogs.msdn.microsoft.com/dotnet/p/targetingdotnet/ which makes it relatively easy to install the required version. But can't reproduce even that now -- not sure what changed.

Anyway installing .NET 4 by following this link http://dotnetsocial.cloudapp.net/GetDotnet?rClrVer=v4.0 was a better experience for me than using the nsis installer built from this PR -- mainly because I could see the progress, was prompted to close running .NET programs and could avoid unexpected restarts.

mattock:
I removed .NET framework from my Windows 7 laptop and placed chipitsine's _OpenVpnServ2.exe_ and _OpenVpnServ2.exe.config_ into OpenVPN's bin folder. Double-clicking on OpenVpnServ2.exe did bring up the _"You need to install .NET framework 4.0 or greater"_ pop-up window.

That said, I believe the usual way to start OpenVPNService is to

* Start the service from the mmc.exe snap-in
* Use "net.exe start openvpnservice"
* Use "sc.exe start openvpnservice"
* Use Powershell "Start-Service" CmdLet (assuming that is possible with .NET 3.5)  

In all of these cases one gets a very obscure error message with no hint of a missing .NET 4.0 installation.

I think a fair compromise would be to detect if .NET 4.0 is missing, and to offer to open a reasonable .NET 4.0 download page using the default browser. This would be roughly analogous to what we do when a Windows XP user tries to install OpenVPN 2.4.x. So the .NET install would not be automatic, and the user would need to take care of it by him/herself outside of the OpenVPN installer.

Does this sound reasonable?

selvanair:
Looks like the kind of popup shown when the required .NET version is not found depends on some helper packages on the system.
Anyway, before starting  the service, it has to be installed and "openvpnsevr2 -install" will fail with the popup if .NET 4.0 or higher is not found. Yet its not ideal. The biggest issue I had with the PR can cause unexpected restarts with no prompts and the installation does not continue after restart. So there are a couple of options:
1. Improve installation within the nsis installer by downloading a minimal online installer (would be only 1 or 2 MB) and running it with /norestart or /promptrestart (untested, may need considerable testing to get right, and we'll have to ask the user to restart manually if needed..)
2. Install the service like:
sc create OpenVPNService /binPath= path_to_openvpnserv2 /depend= tap0901/dhcp 
instead of 'openvpnserv2 -install' which would fail without .NET 4. Then point the user to a .NET install page at the end of the installation.
3. Have an self-contained installer for openvpnserv2 and call it from the openvpn installer

Option (2) is essentially what @mattock is suggesting and may be easiest to implement? We can look into (1) and (3) for a later release.

As the automatic service is for somewhat advanced users, requiring additional steps to run it is not that taxing for users.

mattock:
@selvanair : option 1 (download and run minimal .NET installer) sounds reasonable, if it is opt-in, and if we can find a really stable link for the .NET minimal installer. We would also need to ensure that the .NET install works reliably, so that we don't get flogged for problems with Microsoft's code/servers.

I would prefer to avoid option 3 (separate openvpnserv2 installer), as that would lead to redundancy and surprising issues with passing around data (e.g. variables) between the installers. For example, we'd need to cover the registry key handling in the OpenVPN installer (for OpenVPN-GUI) as well as in openvpnserv2 installer (for openvpnserv2).

I agree that right now option 2 (install with sc, link to .NET download page at the end) seems reasonable right now.

chipitsine:
links are tricky within nsis. it is somewhat related to nsDialog.
I'd stay with messagebox (i.e. no link)

I'll test "intermediate commit", if it is ok, I'll rewrite "installation" magic to "sc.exe"

selvanair:
As an alternative I have opened a PR -- its small commit on top of this PR so could be squashed to this one if found acceptable  (its rebased on to  PR 58 assuming we'll want to merge that one first).

The idea is to just delay the installation of NET to the end and do it in a standalone mode to avoid surprises (unexpected restarts) and show the user what's going on. Please test.

mattock:
This was merged as part of PR https://github.com/OpenVPN/openvpn-build/pull/60.

PR #55, opened 2016-11-25
Use active-setup registry entry to add GUI to startup menu of users

selvanair:
- This adds an active-setp entry in HKLM which will add the required
  HKCU key for automatic launching of the GUI on startup
- The HKCU entry for a user will only be added once when the user logs in
  after a new installation
- As the startup key is in HKCU, the user can change it without admin rights
- For silent installs use SELECT_LAUNCH=0/1. It is selected by default

### DIFF
--- a/windows-nsis/openvpn.nsi
+++ b/windows-nsis/openvpn.nsi
@@ -127,6 +127,7 @@ LangString DESC_SecAddShortcuts ${LANG_ENGLISH} "Add ${PACKAGE_NAME} shortcuts t
 
 LangString DESC_SecFileAssociation ${LANG_ENGLISH} "Register ${PACKAGE_NAME} config file association (*.${OPENVPN_CONFIG_EXT})"
 
+LangString DESC_SecLaunchGUIOnLogon ${LANG_ENGLISH} "Launch ${PACKAGE_NAME} GUI on user logon."
 ;--------------------------------
 ;Reserve Files
   
@@ -322,6 +323,9 @@ Section /o "TAP Virtual Ethernet Adapter" SecTAP
 	WriteRegStr HKLM "SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\${PACKAGE_NAME}" "tap" "installed"
 SectionEnd
 
+Section /o "Launch ${PACKAGE_NAME} GUI on User Logon" SecLaunchGUIOnLogon
+SectionEnd
+
 Section /o "${PACKAGE_NAME} GUI" SecOpenVPNGUI
 
 	SetOverwrite on
@@ -337,6 +341,21 @@ Section /o "${PACKAGE_NAME} GUI" SecOpenVPNGUI
 		CreateDirectory "$SMPROGRAMS\${PACKAGE_NAME}"
 		CreateShortCut "$SMPROGRAMS\${PACKAGE_NAME}\${PACKAGE_NAME} GUI.lnk" "$INSTDIR\bin\openvpn-gui.exe" ""
 		CreateShortcut "$DESKTOP\${PACKAGE_NAME} GUI.lnk" "$INSTDIR\bin\openvpn-gui.exe"
+	${EndIf}
+
+	; Using active setup registry entries to set/unset GUI to launch on logon for each user.
+	; If the user removes the GUI from startup items it will not be re-added or removed on subsequent
+	; installs unless the value of "Version" is updated (do this only if/when really necessary).
+	; Ref: https://helgeklein.com/blog/2010/04/active-setup-explained/
+	WriteRegStr HKLM "Software\Microsoft\Active Setup\Installed Components\${PACKAGE_NAME}_UserSetup" "" "OpenVPN Setup"
+	WriteRegStr HKLM "Software\Microsoft\Active Setup\Installed Components\${PACKAGE_NAME}_UserSetup" "Version" "2,4,0,0"
+	WriteRegDword HKLM "Software\Microsoft\Active Setup\Installed Components\${PACKAGE_NAME}_UserSetup" "IsInstalled" 0x1
+        ; DontAsk = 2 is used to not prompt the user
+	WriteRegDword HKLM "Software\Microsoft\Active Setup\Installed Components\${PACKAGE_NAME}_UserSetup" "DontAsk" 0x2
+	${If} ${SectionIsSelected} ${SecLaunchGUIOnLogon}
+		WriteRegStr HKLM "Software\Microsoft\Active Setup\Installed Components\${PACKAGE_NAME}_UserSetup" "StubPath" "reg add HKCU\Software\Microsoft\Windows\CurrentVersion\Run /v OPENVPN-GUI /t REG_SZ /d $\"$INSTDIR\bin\openvpn-gui.exe$\" /f"
+	${Else}
+		WriteRegStr HKLM "Software\Microsoft\Active Setup\Installed Components\${PACKAGE_NAME}_UserSetup" "StubPath" "reg delete HKCU\Software\Microsoft\Windows\CurrentVersion\Run /v OPENVPN-GUI /f"
 	${EndIf}
 SectionEnd
 
@@ -494,6 +513,7 @@ ${EndIf}
 	!insertmacro SelectByParameter ${SecOpenVPNEasyRSA} SELECT_EASYRSA 0
 	!insertmacro SelectByParameter ${SecAddPath} SELECT_PATH 1
 	!insertmacro SelectByParameter ${SecAddShortcuts} SELECT_SHORTCUTS 1
+	!insertmacro SelectByParameter ${SecLaunchGUIOnLogon} SELECT_LAUNCH 1
 	!insertmacro SelectByParameter ${SecOpenSSLDLLs} SELECT_OPENSSLDLLS 1
 	!insertmacro SelectByParameter ${SecLZODLLs} SELECT_LZODLLS 1
 	!insertmacro SelectByParameter ${SecPKCS11DLLs} SELECT_PKCS11DLLS 1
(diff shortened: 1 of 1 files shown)


selvanair:
Added a commit to have HKCU startup entries removed on uninstall

chipitsine:
Can you, please, add link to active-setup article to nsi?

I haven't heard of such things before

selvanair:
@chipitsine : There are a few of blogs on, but haven't seen an msdn doc.  Can add this link (https://helgeklein.com/blog/2010/04/active-setup-explained/) when ready to squash (on ACK).

chipitsine:
any blog can disappear.
it is seen on several msdn blogs (which also disappear after some time)

https://www.google.ru/search?q=active+setup+site:msdn.microsoft.com

I do not mind, let it be that blog

selvanair:
@chipitsine : Yes, there are many links, but many are incomplete and/or linked to some installation framework. The link I posted is the best I could find. For links that may not disappear, Wikipedia has a short description though it doesn't explain much.

mattock:
The code and the basic functionality looks good, although we definitely want this tested well before embedding this into release installers. Helge mentions [in his blog post](https://helgeklein.com/blog/2010/04/active-setup-explained/) that these active setup scripts are synchronous and can block user login. So there is potential for havoc here, if for whatever reason openvpn-gui.exe hangs on launch. Not sure when/how that could happen. Thoughts?

chipitsine:
we do not add openvpn-gui.exe to active-setup, we add "reg add ... " instead

mattock:
@chipitsine : Good point. Based on cursory look it seems that scripts that user login script [do not block by default](https://technet.microsoft.com/en-us/library/cc975925.aspx), so our current approach should be fairly safe.

selvanair:
As @chipitsine pointed out the "script" is just "reg add somekey ... /f" with so it cannot block. And "/f" makes sure it wont prompt the user with key exists etc. I tested it on Server 2008 and Windows 7. More tests would be nice. Note that the key is added only during the first install (not if re-installed) so if the user removes it (manually or using the GUI menu) after install it wont be added on an update or re-install.

mattock:
As nobody was interested enough to answer [my query](https://www.mail-archive.com/openvpn-users@lists.sourceforge.net/msg03277.html) on openvpn-users, I don't think we'd get independent testers for this. I will do some smoketests myself just in case. Then I think we should just merge this.

mattock:
I tested this fairly extensively using [this installer](http://build.openvpn.net/downloads/snapshots/openvpn-install-2.4_beta2-I901-active-setup.exe), built from openvpn-build Git "master" plus this PR.

Functionality-vise this PR does what it promised, and the initial setup of the registry looks correct. However, the uninstaller section does not seem to work correctly. I did several "install -> check registry -> logout -> login -> check registry" and "uninstall -> check registry -> logout -> login -> check registry" cycles, and these registry values did not change at all:

* HKLM\Software\Microsoft\Active Setup\Installed Components\OpenVPN_UserSetup IsInstalled
* HKLM\Software\Microsoft\Active Setup\Installed Components\OpenVPN_UserSetup StubPath

This is odd, because the code in the Uninstall section of openvpn.nsi should "just work". As a test I even modified openvpn.nsi to delete the two registry values entirely before writing new content into them, but that did not make any difference. To ease testing, I used this tiny PowerShell script which displays the contents of the relevant registry keys:
```
﻿Get-ItemProperty 'HKLM:\SOFTWARE\Microsoft\Active Setup\Installed Components\OpenVPN_UserSetup'
Get-ItemProperty 'HKCU:\SOFTWARE\Microsoft\Active Setup\Installed Components\OpenVPN_UserSetup'
Get-ItemProperty 'HKCU:\SOFTWARE\Microsoft\Windows\CurrentVersion\Run'
```
So the "reg add" command will remain in StubPath in HKLM forever once it gets there.This alone would prevent _HKCU:\SOFTWARE\Microsoft\Windows\CurrentVersion\Run_ from being modified. Also, based on [Helge's blog post](https://helgeklein.com/blog/2010/04/active-setup-explained/) the StubPath command in HKLM (not mirrored to HKCU, apparently) would not run after the first login, as on subsequent logins the _OpenVPN_Usersetup_ component is already present in HKCU.

Can someone else test the installer linked to above and verify that the results are the same as mine? Or maybe provide an alternative installer that does what this PR promises? Typically odd issues like this mean that I'm overlooking something obvious :smile:.

selvanair:
As you pointed out once the HKLM key OpenVPN_Setup gets copied to HKCU the "reg add .." command will not run again on subsequent user login. Thus user customizations are preserved. 

>So the "reg add" command will remain in StubPath in HKLM forever once it gets there.This alone would prevent HKCU:\SOFTWARE\Microsoft\Windows\CurrentVersion\Run from being modified.

As OpenVPN_Setup key with the hard coded version of 2,4,0,0 is mirrored to HKCU, the "reg add .." oin HKLM will not run again, so user is free to modify HKCU\...\Run\OPENVPN-GUI. User changes will stick. I tested this by changing to not launch from the GUI menu and l on re-login the setting was preserved and the gui did not start automatically.

As for the uninstall logic:

When IsInstalled is changed to 0, on next login, one should see a prompt saying OpenVPN-Setup has been removed and whether you want to remove user settings for it. On clicking yes, it executes StubPath and removes HKCU\...\OpenVPN_Setup.  So setting StubPath to "reg del..." during uninstall should do what we want.

I had tested the uninstall part on vista only manually changing IsInstalled to 0 and StubPath to "red del HKCU\...\Run\OPNVPN-GUI /f" and found it worked. Will test the uninstaller itself.

selvanair:
Tested this again, this time went through the full cycle:

1. Fresh install: 
    checked: all reg keys in place? **OK**
2. Logged out and logged in: GUI started automatically -- wow!  **OK**
  checked:  HKLM\Software\Microsoft\Active Setup\Installed Components\OpenVPN_UserSetup mirrored to HKCU ? **OK**
2. Started the GUI, unchecked the box for "Launch on Windows Startup" (sic)
3. Logged out and Logged in again: GUI did not start. **OK**
4. Unistalled from All Programs->OpenVPN->Unsintall menu: **OK**
    checked: 
   Software\Microsoft\Active Setup\Installed Components\OpenVPN_UserSetup\IsInstalled = 0
  Software\Microsoft\Active Setup\Installed Components\OpenVPN_UserSetup\StubPath = reg delete ...
  OK
5. Logged out and Logged in again: 
    A rather long wait with a message saying configuring....:  windows apparently going through a laundry list after noticing that one installed component is gone (that's the price one pays for "user friendliness")
   Pop up window: "OpenVPN Setup has been removed from this system. Do you want to remove your user settings for this component?" or some such
  Clicked Yes (or was it OK?)
 Checked: the mirrored copy of OpenVPN_UserSetup gone? **OK** (so now reinstall will start with a clean slate)
  HKCU\...\Run\OPENVPN-GUI gone? **OK**
  HKLM\Software\Microsoft\Active Setup\Installed Components\OpenVPN_UserSetup still present? **OK**
     with IsInstalled = 0? **OK**
 (it won't trigger again until IsInstalled is changed back to 1)

Still no clue why it doesn;t work for you...

selvanair:
By the way the only windows test machine I have easy access to from work is 32 bit and all tests were done on it. So here is a hunch: the 32/64 combined uninstaller is probably looking is not looking at the 64 bit RegView on that platform.  That would also explain debbie10's report that the "OpenVPN" key in HKLM was not removed on uninstall... @chipitsine ?

To re-test this after a failed reg key removal, be sure to manually remove all relevant keys first -- especially the Active Setup key and the Run-on-login key in HKCU.

selvanair:
@mattock Please checkout the installer [here](https://github.com/selvanair/openvpn-build/releases/tag/v1.0-test)
I suspect the issue is with bitness handling in uninstaller and this should work on 64 bit machines.

chipitsine:
exactly, I used wrong macro in uninstaller

great!

selvanair:
@mattock committed the bitness check fix for Uninstaller
@chipitsine : Did I get it right? (the last commit)

All commits to be squashed at the end..

mattock:
I built [an installer](http://build.openvpn.net/downloads/snapshots/openvpn-install-2.4_beta2-I904-active-setup.exe) with the latest version of this PR, and this time everything worked as intended on 64-bit Windows 7 Pro.

There is one issue, though... when OpenVPN is uninstalled, Windows will (correctly) notify the user 
that OpenVPN-Setup has been removed and offers to remove user settings for it. However, _if_ the user decides _not_ to remove user settings, Windows will bug him/her about it on every subsequent login. Would it be possible to either

* Remember user's choice in this matter
* Just remove the OpenVPN_UserSetup component on login without prompting the user

selvanair:
@mattock That is annoying indeed. We are delaying this to 2.4_rc2, right? Will look into it.

It would be nice to fix the broken bitness check in uninstall before rc2 installer is built. I'll issue a PR -- its a small fix.

selvanair:
> There is one issue, though... when OpenVPN is uninstalled, Windows will (correctly) notify the user
> that OpenVPN-Setup has been removed and offers to remove user settings for it. However, if the user decides not to remove user settings, Windows will bug him/her about it on every subsequent login. Would it be possible to either
> 
> Remember user's choice in this matter
> Just remove the OpenVPN_UserSetup component on login without prompting the user

If we set DontAsk=2 in addition to IsInstalled=0, the active setup key mirrored in HKCU will be silently removed on user login. This is enough to start from a clean slate if/when re-installed at a later time.

The downside is that the command in stubpath will not run at login post uninstall in this case (after install it'll run), so the entry HKCU\Software\Microsoft\Windows\\CurentVersion\Run\OPENVPN-GUI will not be removed. I think that is OK as non-existent programs set to start-on-logon are silently ignored.

It appears microsoft programs such as mediaplayer and IE are installed with DontAsk=2, so there is some precedent. In the absence of any official docs on Active Setup its hard to know whether there is a better way.

mattock:
@selvanair : let's postpone this to OpenVPN 2.4_rc2. I only have an hour or two of working time today, and I still have to squeeze the rc1 release out.

In my earlier tests, when the uninstall section was not working, there were no adverse effects in having OpenVPN-GUI trying to launch on login. So I think it is fine to leave it there.

selvanair:
Rebased to master, squashed all commits and made with following changes --- see also the commit message
(the old branch may be found at [launch-old](https://github.com/selvanair/openvpn-build/tree/launch-old))

The only change in behaviour is that HKCU keys are silently deleted on login after uninstall.
- Add DontAsk=2 for silent operations (especially during uninstall)
- Change the wording "Windows Startup" to "User Logon" for clarity
- StubPath set to delete the startup key for if "launch on logon" is not selected during install
   (This is required overwrite an old StubPath value if re-installed with a new setting.
     Users who have already logged in and customized their setting not affected.)

Tested on Windows 7

mattock:
With the latest changes this seems to work well, so an ACK. Merging.

PR #56, opened 2016-12-01
Check bitness in uninstaller for correct Registry view

selvanair:
This is required to get the registry keys deleted during uninstall on 64 bit platforms.


mattock:
This makes sense indeed. Merging.

PR #58, opened 2016-12-05
Install interactive service and sample configs as a part of the core

selvanair:
- Interactive service is useful for launching openvpn without elevated
  privileges (using GUI or otherwise), so install it as a part of
  openvpn.exe and dependencies.

- Global config and log directories, sample configs and shortcut
  for creating tls-auth key are also generally useful. Set up these
  too on installing openvpn.exe.

- The automatic service (openvpnserv2) is kept optional, selected by
  default.

- Check for running openvpn instances after (not before) removing
  services. This reduces the chance of the installer exiting because
  openvpn is running.

### DIFF
--- a/windows-nsis/openvpn.nsi
+++ b/windows-nsis/openvpn.nsi
@@ -193,6 +193,14 @@ Section -pre
 		Goto guiEndYes
 
 	guiNotRunning:
+		; Stop & Remove previous OpenVPN service
+		DetailPrint "Removing any previous OpenVPN services..."
+		nsExec::ExecToLog '"$INSTDIR\bin\openvpnserv.exe" -remove'
+		nsExec::ExecToLog '"$INSTDIR\bin\openvpnserv2.exe" -remove'
+		Pop $R0 # return value/error/timeout
+
+		Sleep 3000
+
 		; check for running openvpn.exe processes
 		${nsProcess::FindProcess} "openvpn.exe" $R0
 		${If} $R0 == 0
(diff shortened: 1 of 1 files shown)


mattock:
I read through the diff and the patched openvpn.nsi as a whole. I think the changes make sense and I could not spot anything obviously wrong with the implementation. That said, I don't fully understand what the rationale behind adding the CoreSetup function is, as

* It is called only once
* It does similar things as SecOpenVPNUserSpace, i.e installs binaries and docs

I also did some very basic smoketesting ([link to installer](http://build.openvpn.net/downloads/snapshots/openvpn-install-2.4_rc1-I601-openvpn-build-pr58.exe)) and the changes seemed to work ok.

While doing the review/testing I noticed [some unrelated minor issues](https://github.com/OpenVPN/openvpn-build/issues/59) that we should probably fix.

selvanair:
> I don't fully understand what the rationale behind adding the CoreSetup function

@mattock: There is not reason to have that function. To keep the diff small fo reasy review, I just converted the SecService section to that function taking out openvpnserv2 installation and putting it in a new SecService section. We can put everything in that function into SecOpenVPNUserSpace. If you wish I can make the change or you can edit online and add the changes as a commit on top (I think I gave maintainer write permission for the PR branch) to ease additional review/test.

mattock:
@selvanair : let's keep Coresetup for now, and once we're ready to merge combine it to SecOpenVPNUserSpace. I will send a link to the PR#58 installer to  openvpn-devel - if somebody else wishes to test it or do additional reviews all the better. If not, I can give the ACK and we can merge this within a few days.

I'll postpone fixing https://github.com/OpenVPN/openvpn-build/issues/59 until after this PR is in.

selvanair:
@mattock No one else has commented on this. Based on the IRC meeting discussion shall we finalize this (move CoreSetup into SecUserSpace.. ?).  Then we can start the process of further consolidating and limiting options as discussed. Your proposed cleanup (moving up includes etc may be done as the final step?).

PR #60, opened 2016-12-05
Delay .NET 4.0 installation to the end and do it in a standalone mode

selvanair:
A minor patch on top of @chipitsine 's PR 54 (Also rebased to PR 58). So only the third commit is relevant.

This will avoid surprise restarts and give user some idea of what's going on when the .NET installer runs. Otherwise all advantages of PR 54 retained.

Cons: DotNetChecker.nsh is slightly edited -- should we rename it to avoid confusion with the "official" version?

EDIT: test installer is [here](https://github.com/selvanair/openvpn-build/releases/tag/test-net40-v2)


selvanair:
Added a commit to skip NET install during silent install.

mattock:
This OR worked mostly as advertised on Windows 7 Pro 64-bit.
* Silent install worked correctly (no .NET 4.0 download prompts, even if it was missing)
* Refusing to install .NET 4.0 did not cause installer failure
* Accepting to install .NET 4.0 worked, and the .NET installer launched correctly
* If .NET 4.0 was installed, the installer would not ask to install it

There were three issues, though. On first attempt the .NET installer just hanged for several minutes. Cancelling the installation also hanged for some minutes, after which I decided to reboot the computer (which suprisingly did not hang). Of course none of this our fault, but something to keep in mind.

The second problem is more grave: adding OpenVPNService using sc.exe ([line 270](https://github.com/OpenVPN/openvpn-build/pull/60/files#diff-249dbce533c5e00daaf781d19f17281cR270)) consistently failed on every install.

Third problem is that the .NET 4.0 client libraries that the OpenVPN installer wants to install don't seem enough to run openvpnserv2.exe. After installing the client libraries running "openvpnserv2.exe -install" failed with no error message. When I manually installed the full .NET 4.0 framework fixed "openvpnserv2.exe -install" started working again.

I will look into this today and hopefully provide a fix.

mattock:
@selvanair : how is DotNetChecker.nsh modified? I think we should import the original file first, then, as a separate commit, make our modifications. This way we can easily see the changes we made when we have to upgrade DotNetChecker.nsh. Renaming the file would not be a bad idea, either.

mattock:
The cancelled .NET installs seem to have polluted the root of my C:\ drive with more than dozen temporary install directories. Something to keep in mind :smile:

selvanair:
@mattock: DotNetChecker edits are included in the same commit  (61c16da..) where the install is moved to the end. We can split that into a separate commit. 
The issues you noticed:
1. We cant do much about the NET installer hanging, can we..
2.  sc.exe fails to install: my guess:  if you removed .NET when openvpnserv2 was installed, the service will remain installed. Then while running the installer openvpnserv2 -remove would fail to remove it (no NET), so will sc fail to install it (already installed). So I think this is an artifact of the way we test it. Could you please try first manually remove the service, then NET 4.0
3. Installing client profile appeared to work in my tests -- we can change that to Full. Just need to change 40Client to 40Full in the installer as @chipitsine had it originally.

selvanair:
@mattock: just noticed the edits on your comments: So all look good with NET version changed to '40Full' except for possible nastiness of NET installer hanging and the minor 'service already exits' error from sc which I think is acceptable.

mattock:
@selvanair: yes, I agree everything is now in order. I don't insist on having the DotNetChecker.nsh modifications in a separate commit. I just want that our customizations are easily traceable. Some of the commits could possibly be squashed - your call.

Besides that, an ACK from me, and feel free to merge :+1:

selvanair:
@mattock: This includes the "move interactive service to core" (the first commit) which was done to avoid rebasing, assuming that one will get merged first. So do we merge both together here or separate it out? 

Anyway, I'll squash the last two commits and rename DotNetChecker.nsh for clarity. Will keep chipitsine's original commits as is: it adds the original DotNetChecker.nsh and shows how to use it.

You have to do the honors of merging as I don't have write access (which is perfectly fine).

mattock:
I think this PR is good to go. While the "move interactive service to core" commit does not logically belong to this PR, I'm not overly pedantic about that. It is well isolated and the commit message is clear enough. Merging.

PR #63, opened 2016-12-09
Simplifiy user choices and always install openvpn, dlls and services

selvanair:
As discussed in the IRC meeting of Dec 7 and under Issue 59:

- Install "user-space", dlls, iservice by default
- Service (openvpnserv2), GUI, TAP can be selected/unselected
  (selected by default)
- EasyRSA is optional (unselected by default)
- Add-to-path, add-shortcuts, launch-on-logon and file-asociation
  grouped into advance options (selected by default)
- If Service is not selected, openvpnserv2 is not removed, only stopped

It is now possible to unselect service, GUI, TAP in any combination
and still have a working setup using previously installed versions
of those components. However, lack of an installed version of TAP will
cause services not to startup -- this is not checked or warned against.

All hidden and displayed sections may be selected/unselected from the
command line but consistency of those combinations is not guaranteed.

### DIFF
--- a/windows-nsis/openvpn.nsi
+++ b/windows-nsis/openvpn.nsi
@@ -200,9 +200,12 @@ Section -pre
 
 	guiNotRunning:
 		; Stop & Remove previous OpenVPN service
-		DetailPrint "Removing any previous OpenVPN services..."
+		DetailPrint "Removing any previous OpenVPN interactive service..."
 		nsExec::ExecToLog '"$INSTDIR\bin\openvpnserv.exe" -remove'
-		nsExec::ExecToLog '"$INSTDIR\bin\openvpnserv2.exe" -remove'
+		Pop $R0 # return value/error/timeout
+		DetailPrint "Stopping OpenVPN automatic service (may fail if not found)..."
+		; Only stop now, will be removed in SecService if selected
+		nsExec::ExecToLog '"$SYSDIR\net.exe" stop OpenVPNService /yes'
 		Pop $R0 # return value/error/timeout
 
 		Sleep 3000
@@ -228,8 +231,15 @@ Section /o "-workaround" SecAddShortcutsWorkaround
 	; as we don't want to move SecAddShortcuts to top of selection
 SectionEnd
 
-Section /o "${PACKAGE_NAME} User-Space Components" SecOpenVPNUserSpace
+Section /o "-launchondummy" SecLaunchGUIOnLogon0
+	; this section should be selected as SecLaunchGUIOnLogon
+	; this is here as we don't want to move that section to the top
+SectionEnd
 
+; We do not make this hidden as its informative to have displayed, but make it readonly (always selected)
+Section "${PACKAGE_NAME} User-Space Components" SecOpenVPNUserSpace
+
+	SectionIn RO ; section cannot be unchecked by user
 	SetOverwrite on
 
 	SetOutPath "$INSTDIR\bin"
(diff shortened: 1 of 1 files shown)


selvanair:
I tested this with several combinations on Windows 7 but more tests needed. Except for service remove moved to SecService, all changes are moving around sections and adding "-" to the names of some to hide. Yet, the changes look pervasive... sigh.. I like the resulting look of the selections page:
installer

selvanair:
@mattock: Form my side this looks good but a review is needed.

mattock:
@selvanair : ok, I'll review the changes.

selvanair:
Resolved the conflicts after merging PR 62. Commits to be squashed after review.

windows-nsis/openvpn.nsi
```diff
@@ -228,8 +231,15 @@ Section /o "-workaround" SecAddShortcutsWorkaround
 	; as we don't want to move SecAddShortcuts to top of selection
 SectionEnd
 
-Section /o "${PACKAGE_NAME} User-Space Components" SecOpenVPNUserSpace
+Section /o "-launchondummy" SecLaunchGUIOnLogon0
```
mattock: Both _SecAddShortcuts_ and _SecLaunchGUIOnLogon_ are in a separate SectionGroup. I'm wondering whether we truly need this trick or not. Any ideas?

selvanair:
I thought that's a harmless "error". Even if the service exists, in most setups it will not be running and net will complain that it cant stop a service that is not started. These messages could be avoided by first running a "sc query .." to see the service exists and is started and skip the stop command if not. But the current approach appears simpler and thus more robust (less things to go wrong..)

windows-nsis/openvpn.nsi
selvanair: This is because nsis does not allow to refer to a section before its defined. We do not want to move these sections up as then they will then get listed at the top. If we hide them we could move them up and get rid of the dummies. Currently I have added them to "Advanced" indicating that a user normally should leave them alone. And we want "Advanced" to appear at the bottom of the options list.

Is there a better way of achieving this?

mattock:
In general the changes look good, even though I added a few line comments. I will build a test installer today and test it tomorrow.

mattock:
Test installer [here](http://build.openvpn.net/downloads/snapshots/openvpn-install-2.4_rc1-I601-pr63.exe). Also includes tap-windows installer with [the external manifest](https://github.com/OpenVPN/tap-windows6/pull/22).

selvanair:
> Test installer here
> <http://build.openvpn.net/downloads/snapshots/openvpn-install-2.4_rc1-I601-pr63.exe>.
> Also includes tap-windows installer with the external manifest
> <https://github.com/OpenVPN/tap-windows6/pull/22>.
>

Tested on a Windows 7 VM:
install on top of a previous one: OK
clean install after uninstall      : OK
re-install with TAP unchecked: OK (installed  TAP preserved)
re-install with service unchecked: OK (installed openvpnserv2 install
preserved (hurrah!), but service stopped[*])
In all cases addtap.bat does popup UAC prompt.

[*] Not ideal, but as expected.

selvanair:
@mattock: is this ready for 2.4.0?

mattock:
@selvanair : I did basic smoketesting on the installer I had linked to, plus tested silent install and uninstall. I could not find any obvious problems with this PR, and I reviewed the changed earlier, so ACK and marge.

PR #65, opened 2017-01-06
added /SOLID

chipitsine:
see http://nsis.sourceforge.net/Reference/SetCompressor for more details

### DIFF
--- a/windows-nsis/openvpn.nsi
+++ b/windows-nsis/openvpn.nsi
@@ -8,7 +8,7 @@
 
 ; OpenVPN install script for Windows, using NSIS
 
-SetCompressor lzma
+SetCompressor /SOLID lzma
 
 !define PRODUCT_PUBLISHER "OpenVPN Technologies, Inc."
 


chipitsine:
can we merge it?

mattock:
@chipitsine : what is the difference in size between SOLID and non-SOLID installers?

The /SOLID flag means that the installer has to be extracted fully before it can be launched. People who have huge installers apparently see very long delay (10-12 seconds) before the installer starts[[1]](http://neilmitchell.blogspot.fi/2013/11/nsis-plugins-and-lzma-solid.html)[[2]](http://forums.winamp.com/archive/index.php/t-208942.html). In our case the delay is probably reasonable (<0.5 secs).

chipitsine:
with         SOLID:  3378698
without    SOLID:  3830845

chipitsine:
in our case, I suspect the delay is due to windows defender (or other antivirus)

mattock:
@chipitsine : do you have a /SOLID-enabled installer available somewhere?

chipitsine:
http://chipitsine.github.io/solid/openvpn-install-2.4.0-I601.exe

mattock:
The above installer displays an "Unpacking data" window for <0,5 seconds on a I5 laptop. I think that is acceptable. About ~12% is cut off the installer size. I think this is a fair compromise, so let's get this in.

PR #69, opened 2017-01-07
quick & dirty fix for trac#806

chipitsine:
* supress any output when stopping automatic service (either net.exe or sc.exe produces a lot of messages)
* always install openvpnserv2.exe using sc.exe (it produces less output and such way seems to be better because it register dependancy)

in russian windows console utilities (such as net.exe, sc.exe) display output in cp866 (as far as I know, no other language is localized in such strange way), so it is displayed as something scrambled in installer GUI.

this PR does not resolve issue, but it reduces "scrambled" output.


chipitsine:
just for the record

- there's no actively maintained "service management" NSIS plugins
- the way of changing command line tools language https://wandersick.blogspot.com/p/change-non-english-command-line.html seems to be very complicated

PR #70, opened 2017-01-08
fix for trac#806

chipitsine:
I will upload installers soon

### DIFF
--- a/windows-nsis/openvpn.nsi
+++ b/windows-nsis/openvpn.nsi
@@ -201,11 +201,11 @@ Section -pre
 	guiNotRunning:
 		; Stop & Remove previous OpenVPN service
 		DetailPrint "Removing any previous OpenVPN interactive service..."
-		nsExec::ExecToLog '"$INSTDIR\bin\openvpnserv.exe" -remove'
+		nsExec::ExecToLog /OEM '"$INSTDIR\bin\openvpnserv.exe" -remove'
 		Pop $R0 # return value/error/timeout
 		DetailPrint "Stopping OpenVPN automatic service (may fail if not found)..."
 		; Only stop now, will be removed in SecService if selected
-		nsExec::ExecToLog '"$SYSDIR\net.exe" stop OpenVPNService /yes'
+		nsExec::ExecToLog /OEM '"$SYSDIR\net.exe" stop OpenVPNService /yes'
 		Pop $R0 # return value/error/timeout
 
 		Sleep 3000
@@ -268,7 +268,7 @@ Section /o "${PACKAGE_NAME} Service" SecService
 
 	SetOverwrite on
 
-	nsExec::ExecToLog '"$INSTDIR\bin\openvpnserv2.exe" -remove'
+	nsExec::ExecToLog /OEM '"$INSTDIR\bin\openvpnserv2.exe" -remove'
 	Pop $R0 # return value/error/timeout
 	SetOutPath "$INSTDIR\bin"
 	; Copy openvpnserv2.exe for automatic service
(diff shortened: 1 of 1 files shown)


chipitsine:
test installer:  https://chipitsine.github.io/oem/openvpn-install-2.4.0-I601.exe

mattock:
Based on discussion in https://community.openvpn.net/openvpn/ticket/806 this PR does the trick as long as the installer is _not_ Unicode.

@chipitsine : does it make sense to merge this now, if Unicode installers are behind the corner?

chipitsine:
it does make sense.

I'm a bit afraid of couple of things here

- I rolled installer using nsis-2.50, should we upgrade build infrastructure to 2.50 ?
- /OEM is not default option, so it might have some side effects, I asked people to test it. maybe we should ask in openvpn-users mailing list as well ?

chipitsine:
this issue will be addressed separately in unicode installer

mattock:
Typically when I ask people to test installers nobody or maybe one guy responds. So, asking on openvpn-users definitely makes sense.

chipitsine:
can you ask or should I ask myself ?

actually, I think after we will invest in automated testing, it will be more simple

mattock:
@chipitsine : you go ahead and ask.

Improving automated testing is always good, but we can't cover all the corner-cases no matter what we do. So some of the problems will only revealed once real people start using our code in production. The Chocolatey Git snapshot repository will make it easier for these people, who are not terribly afraid of OpenVPN breaking, to stay up-to-date.

It would be really nice to create a system which generates Windows installers automatically from each pull request, and puts them to a public place for download. This would probably be a fair amount of work, but definitely doable using GitHub APIs.

chipitsine:
by automating I mean chocolatey in first place. can you contact current chocolatey maintainer ?

as for building installer on every PR, it is 1-2 days of work. jenkins is capable of such github interoperation, I will install it soon. stay tuned.

mattock:
@chipitsine : so Jenkins has a PR-building feature built-in?

chipitsine:
https://wiki.jenkins-ci.org/display/JENKINS/GitHub+pull+request+builder+plugin

jenkins is often crappy, it usually requires some efforts to get it working. but generally speaking it worth that.

mattock:
It seems that recent versions of Buildbot can also handle notifications from GitHub:

* http://docs.buildbot.net/current/manual/cfg-wwwhooks.html#chsrc-GitHub
* https://developer.github.com/v3/activity/events/types/#pullrequestevent

I'll have to upgrade the buildmaster soonish. and will look into PR building at the same time.

selvanair:
Although OEM to ANSI may work for some languages at least for multibyte encodings it will just replace some gibberish with some other gibberish else, isn't it? I don't have a setup to test this so just guessing.  If so, I think its better to wait for the planned unicode version...

chipitsine:
it is better to fix what can be easily fixed instead of waiting.

chipitsine:
I tested on japanese windows.

original installer

original

patched installer (please note "nsis-2.50" on the screen)

oem

selvanair:
Well, that looks good.

chipitsine:
can we merge it?

nobody on list reported a problem:)

mattock:
I think the worst case scenario is that non-english characters in NSI logs look funky for a _different_ subset of people than now. So this is not a high-risk change imho.

What if we do this:
1. We merge this PR
1. I build new 2.4.0 installer with this PR
1. I update 2.4.0 download links to point to the new installer
1. We monitor the mailing list and forums for a while for problems

If problems are found, I can trivially point the download links back to the previous, known-good installer.

chipitsine:
I know of only russian windows to have different code pages 866 <--> 1251 for console and gui. It might happen, there's no any other language like that.

however, I tried japanese (see screenshot above), I can download from MSDN any language and give it a try.

I agree with your steps. Can you merge it and re-release 2.4.0 ?

chipitsine:
good news, https://sourceforge.net/p/nsis/bugs/1171/
so, I expect this to work in nsis-3.02 as well

mattock:
@selvanair : any comments?

selvanair:
@mattock Agreed, let's merge this.

Issue #75, opened 2017-01-16
disable save-passwords


selvanair:
With this feature proposed for the GUI, shall we add the registry key during install (set to 0, i.e., save password enabled by default) and also provide an option to set it to 1 (i.e., save password disabled) ?

Adding the key at install time has advantage that admins need not consult README to see how exactly to spell the key. Providing an option to set/unset during install is also nice, but it adds one more section (though hidden under the Advanced options).

Thoughts?

chipitsine:
Installer option?

mattock:
An installer option / new section would be easier from deployment perspective for admins who don't use configuration management tools (=the large majority).

chipitsine:
this issue might be closed after PR#77 ?

selvanair:
Implemented in PR #77

PR #77, opened 2017-01-17
Add a disable password save section to installer

selvanair:
- Add a registry key "HKLM\Software\OpenVPN\disable_save_passwords" and
  set it to 0 by default. The key is left unchanged if already present.
- The default behaviour may be changed during installation.
  For silent install use /SELECT_DISABLE_SAVEPASS=1 to set the key to 1.
- If the key is present and set to a non-zero value, the installer
  will not change it back to zero: either uninstall before re-install or
  edit the registry directly.

As discussed in #75 


selvanair:
Now this feature is in the GUI, so would be useful to have this merged.

mattock:
Looks reasonable code-vise. I did not test this, but it looks like it "should just work".

Having the registry key under OpenVPN instead of OpenVPN-GUI is a nice touch, as other GUIs can make use of it more easily.

PR #78, opened 2017-01-17
Always overwrite executable path in registry

selvanair:
As the openvpn installation path is always overwritten by the installer, overwriting the exe_path makes sense. Without this some installations will be broken causing the GUI to fail with "CreateProcess openvpn.exe --version" failed etc..


mattock:
I think it makes sense to always overwrite _exe_path_. The rest of the registry entries are "configuration items" that the user may have customized, and should not change at reinstall.

PR #79, opened 2017-01-18
Enable ASLR/DEP by default

mattock:
URL: https://community.openvpn.net/openvpn/ticket/325

### DIFF
--- a/generic/build.vars
+++ b/generic/build.vars
@@ -41,4 +41,10 @@ EXTRA_OPENSSL_CONFIG="${EXTRA_OPENSSL_CONFIG:--static-libgcc}" # uncomment if op
 #EXTRA_PKCS11_HELPER_CONFIG
 #EXTRA_OPENVPN_CONFIG
 
+EXTRA_TARGET_CFLAGS="-Wl,--dynamicbase,--nxcompat"
+
+# You can enable --high-entropy-va for 64-bit builds if you have recent enough
+# mingw-w64
+#EXTRA_TARGET_CFLAGS="-Wl,--dynamicbase,--nxcompat,--high-entropy-va"
+
 TARGET_ROOT="${TARGET_ROOT:-/}"


mattock:
As discussed [here](https://www.mail-archive.com/openvpn-devel@lists.sourceforge.net/msg13935.html), this PR does not truly turn on ASLR. The produced executables and libraries _look_ like they have ASLR enabled, but their location in memory is not randomized at all.

selvanair:
@mattock: I think the libraries are ok. So this PR is still useful as (i) it enables DEP and (ii) openssl and other dlls will get ASLR enabled and which is needed for the GUI. I would say add --high-entropy-va as well (for 64 bit builds only) and merge this.

For openvpn.exe can be handled in a separate patch here or in openvpn depending on how we want to handle it.

mattock:
@selvanair looks like the --high-entropy-va flag is not supported on the Ubuntu 14.04 version of mingw-w64:
````
$ /usr/bin/x86_64-w64-mingw32-ld --help|grep entropy
````
This makes sense as the commit that introduced --high-entropy-va was made in August 2014:

* https://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git;a=commitdiff;h=2d5c3743a3a223dc6467ff022124b4f2f1844c91;hp=d36430db79dc8eccb9cb188e1234f7ff044f79b4

Should we merge this PR as-is, or add --high-entropy-va to _build.vars_, but comment it out by default so that it does not break any builds?

selvanair:
.> Should we merge this PR as-is, or add --high-entropy-va to build.vars, but comment it out by default so that it does not break any builds?

Sounds good to me.

mattock:
@selvanair additional EXTRA_TARGET_CFLAGS with --high-entropy-va is now in generic/build.vars but it is commented out by default.

Any idea if the linker will just ignore --high-entropy-va on 32-bit builds, or if I need to add some extra logic to generic/build?

selvanair:
Yes 32 bit linker will not recognize --high-entropy-va and error out.  Adding this conditionally in a shell script could be tricky, perhaps. May be just define EXTRA_TAGET_LDFLAGS during build along with CHOST and CBUILD?

mattock:
Based on the comments I'm merging this as-is. Further improvements such as those discussed in the previous two comments can be implemented as a separate PR.

Issue #74, opened 2017-01-16
Restoring service states on OpenVPN install?


mattock:
The current [OpenVPN Chocolatey package](https://chocolatey.org/packages/openvpn) contains Powershell code that [checks](https://github.com/wget/chocolatey_package_openvpn/blob/master/tools/chocolateyInstall.ps1#L53) the state of OpenVPN services (stopped, running) and restores it after install. As discussed [here](https://github.com/wget/chocolatey_package_openvpn/issues/2) it would probably make sense for openvpn.nsi to take care of this instead, so that non-chocolatey users would also benefit from this.

Shall I implement this functionality in openvpn.nsi?

chipitsine:
there's a check of running openvpn.exe, which makes a little sense. it seems it should take of "manually started openvpn.exe"

otherwise, we should care of stopping/running services ... only if user checked to reinstall certain service.


all the above makes me sad, it is complicated in either chosen way

mattock:
Looking at openvpn.nsi this is what we do currently:

* _Section -pre:_ stop OpenVPNService and OpenVPNServiceInteractive ([line 204 ->](https://github.com/OpenVPN/openvpn-build/blob/master/windows-nsis/openvpn.nsi#L204))
* _Section SecService:_ uninstall and then install OpenVPNService ([line 208](https://github.com/OpenVPN/openvpn-build/blob/master/windows-nsis/openvpn.nsi#L271))
* _Section -post:_ set OpenVPNServiceInteractive to start on boot, and also launch it ([line 633 ->](https://github.com/OpenVPN/openvpn-build/blob/master/windows-nsis/openvpn.nsi#L633))

Basically the installer resets the state of OpenVPNServiceInteractive on every install (ensure it's enabled and running), and stops OpenVPNService without restarting it. I think this could definitely be improved if we checked the state (Installed? Running?) of the services in -pre section and restoredthose states in -post section. What happens in the middle does not affect the user that much.

wget:
Actually I was making sure both `OpenVPNServiceInteractive` and `OpenVPNService` were reset to the exact state they were before upgrading the chocolatey package. This was requested by an user (and my own use case as well) relying on OpenVPN to establish an RDP connection to his Windows Server. By default as the installer was not resetting the service states, this led to basically cutting the branch on which we were sitting...
+1 for upstream implementation (i.e. in the NSIS installer).

chipitsine:
[...]

PR #80, opened 2017-01-18
Restore service states after install

mattock:
This PR adds NSIS Simple Service Plugin and uses it to

* Get the state of OpenVPNServiceInteractive and OpenVPNService at the beginning of the install
* Restores the states after install (or if installation cannot proceed)
* Sets our sane defaults (enable on boot and start) for OpenVPNServiceInteractive if it was not present (installed) when the install started.

Thus the state of the system remains the same before and after install. This problem was already solved in the OpenVPN Chocolatey package, as discussed in [issue #74](https://github.com/OpenVPN/openvpn-build/issues/74) using Powershell code. It was agreed that implementing this in NSI makes more sense.

The Simple Service Plugin could be used to manage service starting, stopping and configuration also. RIght now we use a mix of "openvpnserv?.exe -install", "net.exe" and "sc.exe" to do service management. I can move all of that to Simple Service Plugin if we want that.

Integration to openvpn.nsi (=this PR) has not yet been tested, but it "should work". Test installers will follow later.


chipitsine:
can you please compare sc.exe / net.exe / ?-install /SimplePlugin from the following point of view ... if you run install/uninstall/install, you can see fails because uninstall deletes services in "pending remove" way. I did not notice which of the method leads to that.

we should choose the method which actually removes instead of marking as "pending remove"

mattock:
@chipitsine : I'm not sure how to reproduce the PENDING REMOVE issue. Do you get that consistently, or just occasionally?

I fixed a typo (SectionEnd vs. FunctionEnd) and force-pushed. I also switched to using SimpleSC commands for most service control stuff. Based on brief testing things still seem to work as expected. A test installer based on current state of this PR is [here](http://build.openvpn.net/downloads/temp/openvpn-install-2.4.0-I602-pr80v2.exe).

I will now look into replacing the "openvpnserv?.exe -install" calls with SimpleSC.

chipitsine:
no, I'm not sure. ok, I'll reproduce by myself later

chipitsine:
there's massive eventlog usage, like this

https://github.com/OpenVPN/openvpnserv2/blob/master/Service.cs#L120

I'm afraid we introduced a bug by installing service by sc.exe (and not creating event channel).


what I suggest, is to change logging to standard "Application" event log, as people look at it more often.
or otherwise we need to create event channel during install as well.

P.S. I didn't try to run installer that way (i.e. install without .net 4.0 installed in order to use sc.exe)

mattock:
@chipitsine : fortunately sc.exe was only used if .NET 4.0 was not present. You say "massive eventlog usage": do you have sample of an event log with the problem?

Using the Application log makes sense imho, as openvpnserv.exe is already logging in there.

I also wonder how this affects the move to SimpleSC...

mattock:
Looking at application event logs I see errors from openvpnserv.exe related to missing registry key. I don't have the full english version of the message, but here's the meat of it:

    Error querying registry value: HKLM\SOFTWARE\OpenVPN\ovpn_admin_group

Example Powershell to get latest full messages:

    Get-EventLog Application -Newest 500|Where-Object { $_.Source -eq "openvpnserv" }|Format-List

This is not related to openvpnserv2, though, but should be fixed regardless.

selvanair:
The main difference between using sc and net is that  "net stop", "net start" etc waits for the sate change to complete while sc returns immediately. The SimpleSC plugin will wait (with a timeout) so that way its like using net.exe which is the behaviour we want. 

The issue of "this service has been marked for deletion" (I suppose that's what @chipitsine is referring to?) appear to happen sporadically no matter net is sc is used.  I think its almost always related to something holding up the service from getting fully deleted -- could be task manager, mmc or some such. But not 100% sure.

chipitsine:
no, I meant that openvpnserv2 is using different event log, not "Application" by default. it is using it's own log, somewhat like "OpenVpnServ", I do not recall exact name

mattock:
More commits added. Test installer v4 [here](http://build.openvpn.net/downloads/temp/openvpn-install-2.4.0-I602-pr80v4.exe).

selvanair:
> Looking at application event logs I see errors from openvpnserv.exe related to missing registry key. I don't have the full english version of the message, but here's the meat of it:
> 
> Error querying registry value: HKLM\SOFTWARE\OpenVPN\ovpn_admin_group

I'm aware of this error -- its logged only once at service startup and is caused by the fact that I used existing code in the service to read this registry key. The default behavior of the registry parser in the service is to log an error if a key is not found.

I think the best fix is to add this registry key during installation. That also makes it easier for anyone who wants to customize the group name. 

Apart from that event log entries for iservice appear as from unknown source making it difficult to read the log. Fixing this would be nice.

selvanair:
> no, I meant that openvpnserv2 is using different event log, not "Application" by default. it is using it's own log, somewhat like "OpenVpnServ", I do not recall exact name

If that's the case I too think we should change it to log to Application log.

windows-nsis/openvpn.nsi
```diff
@@ -244,11 +244,12 @@ Section -pre
 
 		; Stop & Remove previous OpenVPN service
 		DetailPrint "Removing any previous OpenVPN interactive service..."
-		nsExec::ExecToLog /OEM '"$INSTDIR\bin\openvpnserv.exe" -remove'
+		SimpleSC::RemoveService OpenVPNService
```
selvanair: Should be "SimpleSC::RemoveService OpenVPNServiceInteractive" isn't it ? 
OpenVPNService has to be removed in the secService if selected.

windows-nsis/openvpn.nsi
```diff
@@ -247,7 +247,7 @@ Section -pre
 
 		; Stop & Remove previous OpenVPN service
 		DetailPrint "Removing any previous OpenVPN interactive and legacy services..."
-		SimpleSC::RemoveService OpenVPNService
+		SimpleSC::RemoveService OpenVPNServiceInteractive
```
selvanair: Ah, then ignore my previous comment :)

selvanair:
One thing seems to be missing: restoring the service starttype. If the user had OpenVPNService set to autostart after upgrade it will revert to manual unless the service starttype is cached and restored. Same dor interactive service  -- if anyone has it set to manual (though they shouldn't)  it will get reverted to auto unless we restore the starttype.

That said, why are we removing the interactive service before upgrade -- just stopping should be enough and then the starttype would be preserved. Same for legacy service. For OpenVPNService we do have to remove it before upgrade as the existing one may be the legacy service left behind from old versions.

chipitsine:
I did have a look, actually it is "Application" log. OpenVpnService register new event source

it says "Creating EventLog source OpenVpnService in log Application ..."

we may either move that to "service startup code", i.e. service will perform current installation action (including event source creation) during startup, or we can run something like that during nsis

```
eventcreate /so OpenVpnService /t INFORMATION /id 1 /d OpenVpnService /L Application
```

selvanair:
> e may either move that to "service startup code", i.e. service will perform current installation action (including event source creation) during startup, or we can run something like that during nsis
> 
> ```eventcreate /so OpenVpnService /t INFORMATION /id 1 /d OpenVpnService /L Application```

Is this an nsis plugin? If you are referring to eventcreate.exe in windows, that will just write a one-time event to the event log.

mattock:
I refactored the service handling code quite a bit in the last commit. A test installer based on latest code is [here](http://build.openvpn.net/downloads/temp/openvpn-install-2.4.0-I602-pr80v5.exe).

The new behavior In a nutshell:
* _OpenVPNServiceLegacy_ and _OpenVPNServiceInteractive_ are no longer uninstalled, only stopped
* _OpenVPNService_ is removed as it might refer to OpenVPNServiceLegacy if upgrading from OpenVPN 2.3.x. 
* State of all previously existing services is always restored
* If a service did not exist before, it gets created with default values, and in the case of _OpenVPNServiceInteractive_, started.
* Service state in this context includes its _status_ (running, stopped) and its _startmode_ (manual, automatic, disabled)

Basically the only ways you should get the new NSI code to change service state is to
* Change a service's startupmode and/or status
* Uninstall OpenVPN (removes all services)
* Install OpenVPN (service no longer exists -> service state reset to default)

I also wrote a [Powershell test script](https://github.com/mattock/openvpn-windows-test/blob/master/Test-Installer.ps1) ([pull request](https://github.com/OpenVPN/openvpn-windows-test/pull/6)) that can be used to ensure that this PR works as intended.

windows-nsis/openvpn.nsi
```diff
@@ -344,11 +412,11 @@ Function CoreSetup
 	!insertmacro WriteRegStringIfUndef HKLM "SOFTWARE\${PACKAGE_NAME}" "priority"    "NORMAL_PRIORITY_CLASS"
 	!insertmacro WriteRegStringIfUndef HKLM "SOFTWARE\${PACKAGE_NAME}" "log_append"  "0"
 
-	; install openvpnserv.exe as a services
 	DetailPrint "Installing OpenVPN Interactive Service..."
-	nsExec::ExecToLog /OEM '"$INSTDIR\bin\openvpnserv.exe" -install'
+	SimpleSC::InstallService "OpenVPNServiceInteractive" "OpenVPN Interactive Service" "32" "3" "$INSTDIR\bin\openvpnserv.exe" "tap0901/dhcp" "" ""
```
selvanair: So we let this error if the service is already present?

windows-nsis/openvpn.nsi
```diff
@@ -344,11 +412,11 @@ Function CoreSetup
...
 
-	Pop $R0 # return value/error/timeout
+	DetailPrint "Installing OpenVPN Legacy Service..."
+	SimpleSC::InstallService "OpenVPNServiceLegacy" "OpenVPN Legacy Service" "32" "3" "$INSTDIR\bin\openvpnserv.exe" "tap0901/dhcp" "" ""
```
selvanair: Same here..

windows-nsis/openvpn.nsi
```diff
@@ -175,6 +175,89 @@ ReserveFile "install-whirl.bmp"
...
+	${EndIf}
+
+	${IfNot} $legacy_service_existed == 0
+		DetailPrint "Setting starttype of OpenVPN Legacy Service to manual"
+		SimpleSC::SetServiceStartType "OpenVPNServiceLegacy" 3
+	${EndIf}
+
+	${IfNot} $service_existed == 0
+		DetailPrint "Setting starttype of OpenVPN Service to manual"
+		SimpleSC::SetServiceStartType "OpenVPNService" 3
```
selvanair: This too looks redundant as its set during install

windows-nsis/openvpn.nsi
```diff
@@ -175,6 +175,89 @@ ReserveFile "install-whirl.bmp"
...
+	; Use default settings for services which were not installed before
+	${IfNot} $iservice_existed == 0
+		DetailPrint "Enabling and starting OpenVPN Interactive service" 
+		SimpleSC::SetServiceStartType "OpenVPNServiceInteractive" 2
+		SimpleSC::StartService "OpenVPNServiceInteractive" "" 5
+	${EndIf}
+
+	${IfNot} $legacy_service_existed == 0
+		DetailPrint "Setting starttype of OpenVPN Legacy Service to manual"
+		SimpleSC::SetServiceStartType "OpenVPNServiceLegacy" 3
```
selvanair: Same here: if not existed, starttype is already set during install. But reinstall to new location requires a binary_path reset.

windows-nsis/openvpn.nsi
```diff
@@ -175,6 +175,89 @@ ReserveFile "install-whirl.bmp"
...
+	Pop $0
+	Pop $service_was_running
+FunctionEnd
+
+Function RestoreServiceState
+	; Use default settings for services which were not installed before
+	${IfNot} $iservice_existed == 0
+		DetailPrint "Enabling and starting OpenVPN Interactive service" 
+		SimpleSC::SetServiceStartType "OpenVPNServiceInteractive" 2
+		SimpleSC::StartService "OpenVPNServiceInteractive" "" 5
```
selvanair: The service StartType in this case is already set by the call to install, so is not needed ?

If the user is installing to a new INSTDIR without uninstalling the previous one we also have to reset the service executable path. So need an ${Else}  SimpleSC::SetServiceBinaryPath .... ?

windows-nsis/openvpn.nsi
mattock: Yes, we can do that. SimpleSC does not produce any output on error, so this is harmless.

mattock:
I'll look into the redundant StartType and binary path issues early next week.

mattock:
I refactored the code quite heavily and squashed all the commits together. I used the Powershell script abaove to ensure that the logic works as expected. I believe all of Selvanair's concerns have been addressed now.

I also found and fixed a bug that was masked before. The problem was that _running_ services would not get removed in the Uninstall section by SimpleSC::RemoveService. These services would then be left behind to cause odd side-effects during subsequent installs.

A test installer is available [here](https://build.openvpn.net/downloads/snapshots/openvpn-install-2.4.0-I602-pr80v18.exe). Let me know if further changes are required.

selvanair:
Looks good and works as promised. One question:
Any reason why  interactive service gets only 5 sec timeout for start/stop while the other two are given 10 secs? As all three are stopped together, cumulatively it seems to get ample time to stop, so just being curious..

ACK.

mattock:
@selvanair : I assumed that OpenVPNService and OpenVPNServiceLegacy need more time as they have to shut down all the OpenVPN instances they're managing. That said, I don't know if all the OpenVPN instances have to actually be killed before the service is marked as "stopped". I'll increase the start/stop timeout for OpenVPNServiceInteractive to 10 seconds, force-push and then merge this.

chipitsine:
how those timeouts are implemented in SimpleSC ?
does it wait for 10 sec and checks for service state ? or does it check every second ?

mattock:
Looks like SimpleSC polls the service status every 250 milliseconds. These polls continue until the desired status is reached. It exists with ERROR_SERVICE_REQUEST_TIMEOUT if timeout is reached before this happens.

chipitsine:
wow

mattock:
Right now the exit code of SimpleSC commands are not displayed in the install log, which could be problematic when diagnosing problems. I'll see if this could be fixed without adding tons of new code.

mattock:
I increased OpenVPNServiceInteractive start/stop timeout to 10 seconds.

Printing exit codes of the SimpleSC commands would increase patch size considerably. I'll merge this as-is, and we can add debugging to places where people encounter issues (if any).

wget:
Sometimes, I wonder if that wasn't easier to leverage msi Windows Installer files. Have just read tutorials, this isn't too complicated to code.

mattock:
@wget : I don't particularly like NSIS as it feels like a big mess. That said, it works very reliably, and more importantly has never prevented us from doing what we need. Most of what our installer does is basic stuff (e.g. files and registry keys) , but based on a quick look there are some exceptions:

* Offers to install .NET 4.5 framework if it is missing (unless install is silent)
* Caches/restores service states
* Refuses to start the install if (manually launched) openvpn.exe instances are running
* Runs the tap-windows6 installer from within the openvpn installer with correct options
* (Injects the publisher certificate to the certificate store)

These boil down to running some commands, processing command output and acting accordingly if necessary. Any ideas how such things would be handled using MSI?

chipitsine:
I beleive such things could be handled by pull requests :)
and, it looks that there's new developer in openvpn area!

wget:
@mattock  +1 NSIS for the mess. Btw, msi offers the advantage to be deployed using the Enterprise Windows Update and using GPO. This is why is should be privileged on Windows environments.

Without being 100% certain, I think msi could meet the aforementioned requirements.

I'm gonna look at this as soon I catched up with the other chocolatey packages I manage (the purpose is to merge them in [coreteampackages](https://github.com/wget/chocolatey-coreteampackages/)) and have finished the code required to have a snapshot of the OpenVPN Chocolatey package (cf. [issue#1](https://github.com/wget/chocolatey-package-openvpn/issues/2)). I begin to become fluent with AppVeyor, but this take time as my free time is limited :)

PR #82, opened 2017-03-15
Resolve snapshot build

chipitsine:
@mattock , can you, please enable "cron job" for openvpn-build repo ?
we missed broken snapshot compilation


chipitsine:
I will investigate it a bit more

PR #88, opened 2017-05-16
Clarify that the certificate management scripts refer to EasyRSA 2

mattock:
From user perspective it was not clear what certificate management scripts the installer provided. Even if the user knew the installer contained EasyRSA, there was no clear indication of whether it was EasyRSA 3 (=easy-rsa on our GitHub page) or EasyRSA 2 (=easy-rsa-old).


mattock:
@selvanair @chipitsine : can we merge this? The changes are trivial and imho make perfect sense.

selvanair:
Makes sense. Please merge.

chipitsine:
Personally, I prefer to use xca. Changes make sense, lets merge

PR #91, opened 2017-05-25
Updated openSSL binaries version to 1.0.2l

ch1c4um:


mattock:
An identical change was made when OpenVPN 2.4.3 was released, so this PR is no longer needed.

PR #92, opened 2017-06-12
Add a registry entry for ovpn_admin_group set to its default value

selvanair:
This entry is optional as it is only used to override the default
value of "OpenVPN Administrators". However, in its absence, the service
logs an error indicating a missing registry key. Adding the registry entry
during installation prevents the error message and also provides a place
holder for users who want to customize the name of the admin group.

Trac: #892


mattock:
:+1:

PR #93, opened 2017-06-20
Update easy-rsa to 2.3.0

mattock:
The new version uses stronger default hash and encryption algorithms, as well as
a larger DH params size.

Signed-off-by: Samuli Seppänen <samuli@openvpn.net>

PR #94, opened 2017-06-20
Update easy-rsa to 2.3.0

mattock:
The new version uses stronger default hash and encryption algorithms, as well as
a larger DH params size.

This replaces PR https://github.com/OpenVPN/openvpn-build/pull/93.


mattock:
I will merge this with a lazy-ACK to get this into the next installer. The changes to easy-rsa were approved [here](https://github.com/OpenVPN/easy-rsa-old/pull/1).

PR #100, opened 2017-07-17
small fixes and aarch64 support

ordex:
one patch is about adding support for the aarch64 (ARM64) cross-compiler and the others are fixing small glitches I found while using the build script.

### DIFF
--- a/generic/build.vars
+++ b/generic/build.vars
@@ -41,7 +41,7 @@ EXTRA_OPENSSL_CONFIG="${EXTRA_OPENSSL_CONFIG:--static-libgcc}" # uncomment if op
 #EXTRA_PKCS11_HELPER_CONFIG
 #EXTRA_OPENVPN_CONFIG
 
-EXTRA_TARGET_CFLAGS="-Wl,--dynamicbase,--nxcompat"
+EXTRA_TARGET_CFLAGS="${EXTRA_TARGET_CFLAGS:-Wl,--dynamicbase,--nxcompat}"
 
 # You can enable --high-entropy-va for 64-bit builds if you have recent enough
 # mingw-w64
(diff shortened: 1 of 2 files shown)


mattock:
So the error message is this:
```
Build openssl
target already defined - mingw (offending arg: Wl,--dynamicbase,--nxcompat)
FATAL: Configure openssl
```
The arguments it complains about are the defaults given in [generic/build.vars#44](https://github.com/OpenVPN/openvpn-build/blob/master/generic/build.vars#L44). These were introduced fairly recently in [PR#79](https://github.com/OpenVPN/openvpn-build/pull/79).

ordex:
not sure what to do. These flags were also giving me troubles when compiling for arm64 and I had to comment them out. Will try to see if there is something I introduced that is conflicting.

chipitsine:
as for amd64 - we should add it to travis-ci build matrix, I think
otherwise, it could get broken

chipitsine:
we can strip offending commit for now, right ?

ordex:
@mattock when was the last time that these tests have been executed on master?

mattock:
It is interesting that commit [build.vars: allow EXTRA_TARGET_CFLAGS to be specified on command line](https://github.com/OpenVPN/openvpn-build/pull/100/commits/6ae9644c7246115a61f6bb20e4b89ba0eebdd77d) _seems_ to break the build, as the flags it affects have been present [since May](https://github.com/OpenVPN/openvpn-build/pull/79). Looking at commit logs it looks like [moving from OpenSSL 1.0.2k to 1.0.2l](https://github.com/OpenVPN/openvpn-build/commit/41e412912bee6faa1dee871db1a54f4fe13f4b49) might be the actual problem. But then again, I was able to build with OpenSSL 1.0.2l just fine.

@ordex: could you do what @chipitsine suggested and (temporarily) remove 6ae9644c724. If that helps, then we know a bit more.

ordex:
quite weird because I had this problem also during my build (but I thought it was a aarch64 specific problem) thus I introduced  6ae9644 to allow me to unset those flags from command line. Anyway, I am going to push another branch without this commit and make a pull request for that.

ordex:
This pull request won't be merged as the last commit is breaking the OpenSSL build process

PR #101, opened 2017-07-20
fix and aarch64 bis

ordex:
this is the same as antonio/fix-and-aarch64 but without 6ae9644


ordex:
Interesting, something is green :) then that commit is very fishy

ordex:
@mattock want to merge these two commits for now?

mattock:
Yes, looks good now.

PR #103, opened 2017-07-25
replace EnvVarUpdate.nsh with AddToPath.nsh

chipitsine:
according to http://nsis.sourceforge.net/Path_Manipulation
EnvVarUpdate is not considered as a proper tool for PATH update.
Switching to "AddToPath.nsh" will allow to use non patched nsis,
this step is preliminary for the ongoing migration to nsis3


chipitsine:
any news on this ?

selvanair:
Where is this AddToPath.nsh taken from? -- there is an old script with that name which pre-dates EnvVarUpdate.nsh which will have the same limitation. And skimming through the code I do not see how it solves the 1024 chars limit. A header showing the origin/date and a comment on how the limit is avoided would help.

chipitsine:
I took AddToPath.nsh from https://github.com/nmap/ncrack/tree/master/mswin32/nsis
(as a quick look it is the same as in http://nsis.sourceforge.net/Path_Manipulation

I build test installer and performed smoke test - PATH is updated on both install and uninstall.
however, I did not check how it behaves when PATH is bigger than 1024 chars. I'll check soon

yes, I agree on better documentation, will perform an updated patch

chipitsine:
as for "when the limit was avoided" - I think the link to http://nsis.sourceforge.net/Path_Manipulation added to openvpn.nsi should be enough ?

mattock:
I would definitely like to get rid of building special version of NSIS just to support long PATHs, even though [the process](https://community.openvpn.net/openvpn/wiki/PatchingDebs) is relatively straightforward. I'm building Ubuntu Yakkety (16.10) version of NSIS (2.51) as I write this.

mattock:
Oh, and I think we should start experimenting with MSI installers. The [Chocolatey](https://chocolatey.org/) package maintainer was interested in participating in that project, so we could probably offsource some of the work to him. Of course modifications would be needed to openvpn-build as well.

chipitsine:
I did some MSI recently as well. Looks good for me.

chipitsine:
as for "Chocolatey package maintainer" activities, I think we do underestimate the power of community in general. I seen several very interesting things in openvpn forks, we can make more efforts into coordinating those efforts (I'll give links to "interesting" forks soon)

selvanair:
> as for "when the limit was avoided" - I think the link to http://nsis.sourceforge.net/Path_Manipulation added to openvpn.nsi should be enough ?

Are we looking at the same place? Can you provide a direct link to the AddToPath.nsh you posted? The above link points to a warning and recommends to use the code at https://www.smartmontools.org/browser/trunk/smartmontools/os_win32/installer.nsi?rev=4110#L636 as a fix. The smartmontools installer linked there includes an AddToPath which avoids overwriting path by bailing out if its > NSIS_MAX_STRLEN. Using the built-in ReadRegKey etc it wouldn't be possible to exceed that limit.

What is the string length limit of the NSIS build you used for the test? ${NSIS_MAX_STRLEN} will print it.

That said, I like the idea of moving to MSI. In the mean time we could use a recent NSIS release with special build.

selvanair:
@mattock:
> I would definitely like to get rid of building special version of NSIS just to support long PATHs, even though the process is relatively straightforward. I'm building Ubuntu Yakkety (16.10) version of NSIS (2.51) as I write this.

Can we use the special build (with string length = 8192) from NSIS binary release ? (http://prdownloads.sourceforge.net/nsis/nsis-3.02.1-strlen_8192.zip?download)

chipitsine:
@selvanair :  http://nsis.sourceforge.net/Path_Manipulation#The_Functions


also, I did a check (see attached screenshot), I polluted $PATH with 4000+ characters, installer handled it perfectly

chipitsine:
I beleived that I was using vanilla nsis.

however

```
[ilia@localhost spec]$ grep NSIS_MAX_STRLEN *
mingw-nsis.spec:%global sconsopts VERSION=%{version} PREFIX=%{_prefix} PREFIX_CONF=%{_sysconfdir} ZLIB_W32=%{mingw32_sysroot}  SKIPUTILS='NSIS Menu' STRIP_CP=false NSIS_MAX_STRLEN=8192
mingw-nsis.spec:- Add NSIS_MAX_STRLEN=8192 to sconsopts (RHBZ#1090075).
[ilia@localhost spec]$ 
```

I cannot test what will happen after 8192, win2012r2 only allows 4123 symbols.

well, this PR must be wrong than

chipitsine:
seems that we might consider not to add openvpn.exe to PATH

mattock:
Let's just remove openvpn's bin directory from PATH. Those people who depend on that can manually add it back.

chipitsine:
thanks to @selvanair findings, this PR is just wrong.
so, I close it

TinCanTech:
This breaks all easyrsa versions.

mattock:
@TinCanTech : by "this" you're referring to not adding OpenVPN's bin directory to PATH?

TinCanTech:
By "This" I mean: Stop modifying path to avoid problems #105
I am not apportioning blame simply pointing out a knock on effect.
Reported here: https://github.com/OpenVPN/easy-rsa/issues/148

chipitsine:
well, it is a mess actually. sorry for that.

anyway, openvpn installer can modify bundled easy-rsa scripts, right?

TinCanTech:
> openvpn installer can modify bundled easy-rsa scripts

I don't think that is the right way to go when changing easyrsa is simple.

egroeper:
>  > openvpn installer can modify bundled easy-rsa scripts
> 
> I don't think that is the right way to go when changing easyrsa is simple.

I agree with @ecrist. To me it seems, that this should be fixed on OpenVPN-side, not easy-rsa.
What easy change do you suggest @TinCanTech ?

Would bundling an adjusted vars file with correct path to openssl binary (`EASYRSA_OPENSSL`) fix this issue well?
I see possible problems, if someone adjusts his vars file (as expected!) and we possibly overwrite it in case of an OpenVPN update.
The question is: What is the cleanest solution for this issue? Of course we could simply patch the `set_var EASYRSA_OPENSSL openssl` line of easy-rsa script when building OpenVPN installers.

Thinking about it, I would prefer the latter option. Because then user is free to customize his (all) settings in a vars file as he likes.

TinCanTech:
> I agree with @ecrist. To me it seems, that this should be fixed on OpenVPN-side

I respectfully disagree .. see below:

> I see possible problems, if someone adjusts his vars file (as expected!) and we possibly overwrite it in case of an OpenVPN update.

Upgrading openvpn on Windows does **not** over write any user customised files. 
IE. easy-rsa/vars.bat and easy-rsa/pki

> Would bundling an adjusted vars file with correct path to openssl binary (EASYRSA_OPENSSL) fix this issue well?

Yes, I have already done this and it works.
eg: vars.bat.example (easy-rsa-2.2.2)
```
rem OpenVPN will install OpenSSL to the following directory by default
rem due to OpenVPN Installer no longer adding openvpn/bin to PATH we specify
rem the path to OpenSSL here, adjust to your specific system if required.

set openssl_path="%ProgramFiles%\openvpn\bin
```
Subsequently, we must change each call to OpenSSL in the easyrsa `.bat` files
eg: build-ca.bat
```
%openssl_path%/openssl ${etc}
```
The required change to easy-rsa-3.x is even simpler but at this time it is not packaged with the OpenVPN Windows installer so not required here.

> Thinking about it, I would prefer the latter option. Because then user is free to customize his (all) settings in a vars file as he likes.

Indeed.

chipitsine:
what if we just pick openssl path from the registry ?

```
@ECHO OFF &SETLOCAL 
FOR /F "tokens=2*" %%a IN ('REG QUERY "HKEY_LOCAL_MACHINE\SOFTWARE\OpenVPN"') DO set "OpenVPNHome=%%b"
ECHO %OpenVPNHome%​
```

egroeper:
> > I agree with @ecrist. To me it seems, that this should be fixed on OpenVPN-side

> I respectfully disagree

I think we have a misunderstanding here. Supplying a customized vars.bat.example file with OpenVPN is "fixing it on the OpenVPN-side" for me.
Adjusting easy-rsa source to support a path variable, would be fixing it on the easy-rsa side. But honestly I don't use easy-rsa bundled with Windows installers and didn't know, that easy-rsa 2.0 still is bundled. So I only looked at 3.0.

Looking at your suggestions: Wouldn't adding something like `set PATH="%PATH%;<path to openvpn\bin>"` be the easiest solution. We would then not need to touch every batch file.

TinCanTech:
> Looking at your suggestions: Wouldn't adding something like `set PATH="%PATH%;<path to openvpn\bin>"` be the easiest solution. We would then not need to touch every batch file.

That looks better than my idea +1

mattock:
Oh well, I will copy-and-paste my proposal here also - it's already in two other places :smile::

_"I believe a typical way to handle this is to have "EasyRSA command-prompt" shortcut in the Start menu with the correct PATH. That way system path does not have to be modified."_

So basically what @egroeper suggested. The benefit of doing this in the installer ([openvpn.nsi](https://github.com/OpenVPN/openvpn-build/blob/master/windows-nsis/openvpn.nsi)) is that the correct path to OpenVPN is available and we're not forced to use a default value (e.g. _C:\Program Files\OpenVPN\bin_).

chipitsine:
it might be taken on the fly from registry, isn't it just better ?

mattock:
@chipitsine : if we can do that easily then I agree with you.

chipitsine:
@mattock 

vars.bat.example defines openssl_path statically,
we can change it with

```
@ECHO OFF &SETLOCAL 
FOR /F "tokens=2*" %%a IN ('REG QUERY "HKEY_LOCAL_MACHINE\SOFTWARE\OpenVPN"') DO set "openssl_path=%%b"
```

TinCanTech:
> "I believe a typical way to handle this is to have "EasyRSA command-prompt" shortcut in the Start menu with the correct PATH. That way system path does not have to be modified."

Some counter points:
1. EasyRSA (all versions) do **not** currently include "EasyRSA command-prompt" shortcut.
2. EasyRSA installs to protected `%ProgramFiles%\EasyRSA` thus opening a non-elavated command prompt results in "Access Denied" errors.
3. The result of setting "Run as Administrator" to the *proposed* shortcut is that it will **only** open in `%SystemRoot%\system32` **Over riding any location specified by "Start in" with Windows 10**.

Reading `vars.bat.sample` (EasyRSA2x) there is currently **no** provision for specifying openssl path.  @egroeper 's suggestion (*and the solution for which I vote*) does work as expected.  
IE: Add `set PATH="%PATH%;<path to openvpn\bin>"` to `vars.bat.sample`

Reading the `vars.sample` (EasyRSA3x) it is *quite clear* that provision has been made to specify the correct path to openssl. 

Finally, @ecrist has already asked that future `openvpn-intaller-x.exe`'s be shipped with EasyRSA-3x so all this is somewhat moot.  If you don't want to make a song and dance out of it then I suggest EasyRSA2 gets the fix proposed by @egroeper and EasyRSA3x ships as-is.  Although, if EasyRSA3x (as it is now) is installed to `%ProgramFiles%\Openvpn` it will suffer the same problems (1,2,3) as above.

PR #105, opened 2017-08-14
Stop modifying path to avoid problems

egroeper:
In #103 @mattock suggested to remove the functionality to add the OpenVPN binary to the path.
This avoids the problems related to path variables whose content is longer than 1024 characters.

As I didn't see related work, I simply did this (because the problem nags me too).
Don't know who really needs this feature. Seems like it is there for quite a long time.
Is some kind of deprecation needed?

Do we want to keep the PATH-removal stuff around for a while, so that OpenVPN binary gets still removed on uninstallation of existing (updated) OpenVPN installs?


chipitsine:
I'm curious, why does it nag you ?

egroeper:
> I'm curious, why does it nag you ?

**tl;dr**: I'm lazy and tend to forget things.

We are providing OpenVPN to potentially 40k users. To minimize support time and make the use of OpenVPN as easy as possible, we provide preconfigured installation packages. This comes from the time, where embedding additional files into the config file wasn't (fully) possible.
To create this packages, I simply extract the upstream packages, add our files, make our (minor) modifications and repackage everything (run makensis with slightly modified openvpn.nsi). That way we don't rely on signing everything by ourselves. If we don't modify the signed binaries and drivers, we can simply reuse them.

Most of our users don't run into the 1024 chars PATH issue, but some do.
So in the past (a few years ago?) we had requests reporting this issue. So I searched the web and found the docs about the issue and possible workarounds. I compiled my own patched nsis version and installed that as a package. Some time later (a year or more?) it was time to upgrade the linux distribution of the computer where, among other things, our OpenVPN packages are build. Because of that nsis got updated and I simply didn't notice / care / remember. Now, another year later(?), there are again users reporting this issue, I already fixed in the past. So I was providing binaries including this issue for quite some time *again*. 

Now I could rebuild nsis again and wait till I forget about this (again!) or, like stated in #103 (whose discussion I was following), I can remove the PATH stuff and get rid of the issue altogether.
I really doubt, that any of our users want to run OpenVPN manually from commandline or the like.

chipitsine:
wow, we do similar things, however, I'm running something like 1000+ users installation.


as for users, there is no real impact of 1024 chars:

1) there are various builds of makensis, that support 8192 chars (most probably yours is like that)
2) from the user point of view it is not an error. installer tries to modify the PATH and it never returns error here, no matter whether PATH is modified or not.


anyway, it would be kind of you if you would share your experience of provisioning and supporting such large installation (its bigger than mine)

mattock:
Historically when we remove minor feature one or two people who depended on it will complain. But the vast majority of people don't notice anything. Having OpenVPN binaries in PATH only really helps people who run routinely run openvpn.exe manually, and I don't think there are many of them. People who create scripts can just as well use the full path to openvpn.exe. I say let's just get rid of this feature.

mattock:
Based on our discussion in PR https://github.com/OpenVPN/openvpn-build/pull/103 and here I will merge this PR now. This gets rid of the need for a special NSIS build, which simplifies life for everyone building and packaging OpenVPN Windows installers using openvpn-build.

mattock:
This had the unintended side-effect that easy-rsa broke, see https://github.com/OpenVPN/easy-rsa/issues/148. My proposed fix, quoting from the above link:

_"I believe a typical way to handle this is to have "EasyRSA command-prompt" shortcut in the Start menu with the correct PATH. That way system path does not have to be modified."_

PR #110, opened 2017-10-18
Add "PKCS#11 URIs compliant with RFC7512" patch

chipitsine:
added patch from https://github.com/OpenSC/pkcs11-helper/pull/4

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -13,6 +13,6 @@ environment:
       CHOST: x86_64-w64-mingw32
 
 build_script:
-  - "%CYGWIN%\\setup-%ARCH%.exe -g -q -P mingw64-i686-openssl,mingw64-x86_64-openssl,cygwin-devel,unzip,dos2unix"
+  - "%CYGWIN%\\setup-%ARCH%.exe -g -q -P mingw64-i686-openssl,mingw64-x86_64-openssl,cygwin-devel,unzip,dos2unix,patch"
   - "%CYGWIN%\\bin\\bash -lc 'cp `which true` /usr/bin/man2html'"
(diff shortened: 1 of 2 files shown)


mattock:
In [yesterday's community meeting](https://www.mail-archive.com/openvpn-devel@lists.sourceforge.net/msg15655.html) @dazo and @cron2 gave a feature-ACK for bundling this patch in new 2.4 Windows installers.

@chipitsine : does the patch actually apply? I'm asking because the old lzo-2.05 patch had this header:
```
#diff -urNp lzo-2.05/Makefile.am lzo-2.05.1//Makefile.am
```
This PR's patch has this header:
```
diff --git a/lib/pkcs11h-serialization.c b/lib/pkcs11h-serialization.c
```
I think you would need to have this instead:
```
diff --git a/pkcs11-helper-1.11/lib/pkcs11h-serialization.c b/pkcs11-helper-1.11/lib/pkcs11h-serialization.c
```
This way openvpn-build would know that it needs to patch pkcs11-helper.

chipitsine:
@mattock , I tested, patch does apply.

can you please test yourself as well ?

just 'clone -b RFC7512 https://github.com/chipitsine/openvpn-build' and build it ?

chipitsine:
@mattock , how was your test ?

I think, as long as we apply patch with "-p1", the header does not matter

mattock:
@chipitsine : haven't had the time to test yet. That said, I will need to rebuild at least 2.3-based Windows installers late next week. That will be a good time to verify this patch also.

chipitsine:
@mattock , did you test it?

mattock:
As agreed in today's IRC meeting I'll merge this now.

Issue #111, opened 2017-10-20
cross-compiling Win32 build on Ubuntu 16.04 fails if DO_REALLY_STATIC=1


johnmah:
Fails with:

```
...
configure: WARNING: tap-windows.h: accepted by the compiler, rejected by the preprocessor!
configure: WARNING: tap-windows.h: proceeding with the compiler's result
checking for tap-windows.h... yes
checking whether TUNSETPERSIST is declared... no
checking for setcon in -lselinux... no
checking for pam_start in -lpam... no
checking for PKCS11_HELPER... yes
checking for SSL_CTX_new... no
configure: error: openssl check failed
FATAL: Configure openvpn
```

This previously worked for me for OpenVPN 2.3.10 w/ OpenSSL 1.0.2k. Also current master builds fine without DO_REALLY_STATIC variable set. FWIW, also fails on Ubuntu 14.04.

I've attached output for the build process when DO_REALLY_STATIC is set.

[openvpn-build-output.txt](https://github.com/OpenVPN/openvpn-build/files/1402948/openvpn-build-output.txt)

chipitsine:
Thank you for reporting, I'll have a look soon

chipitsine:
it seems, there's  no easy way to get DO_REALLY_STATIC back
it is broken for a long time

you are the only person who complains on that. can please describe your use case  ?

chipitsine:
it does not work even for openvpn-2.3.10 + openssl-1.0.2k

```
git clone -b release/2.3 https://github.com/OpenVPN/openvpn-build.git
cd openvpn-build/generic
IMAGEROOT=`pwd`/image-win32 DO_REALLY_STATIC=1 CHOST=i686-w64-mingw32 OPENSSL_VERSION=1.0.2k OPENVPN_VERSION=2.3.10 OPENVPN_URL=http://openvpn.skbkontur.ru/openvpn-2.3.10.tar.gz CBUILD=x86_64-pc-linux-gnu ./build
....
ssl_openssl.o:ssl_openssl.c:(.text+0x2bcb): undefined reference to `SSL_CTX_free'
ssl_openssl.o:ssl_openssl.c:(.text+0x2bd9): undefined reference to `SSLv23_method'
ssl_openssl.o:ssl_openssl.c:(.text+0x2be1): undefined reference to `SSL_CTX_new'
ssl_openssl.o:ssl_openssl.c:(.text+0x2c25): undefined reference to `SSL_new'
[...]

PR #114, opened 2017-10-26
include missing win32 libraries when using static built openssl

johnmah:
addresses https://github.com/OpenVPN/openvpn-build/issues/111

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,10 @@ matrix:
       env: CHOST=i686-w64-mingw32 SCRIPT=generic/build
     - compiler: ": Win64"
       env: CHOST=x86_64-w64-mingw32 SCRIPT=generic/build
+    - compiler: ": Win32 + static"
+      env: DO_REALLY_STATIC=1 CHOST=i686-w64-mingw32 SCRIPT=generic/build
+    - compiler: ": Win64 + static"
+      env: DO_REALLY_STATIC=1 CHOST=x86_64-w64-mingw32 SCRIPT=generic/build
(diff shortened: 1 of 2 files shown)


chipitsine:
Maybe add here https://github.com/OpenVPN/openvpn-build/blob/master/generic/build.vars#L36 ?

johnmah:
if you move adding the extra libraries required for static linking to the suggested spot, then you will need to duplicate or move the Windows build detection logic into the `build.vars` file (it's done here: https://github.com/OpenVPN/openvpn-build/blob/master/generic/build#L363).

chipitsine:
ok, it does make sense.

however, why do you set instead OPENSSL_LIBS of LDFLAGS?

johnmah:
LDFLAGS implies global flags across all binaries, which may not be the case. OPENSSL_LIBS implies that they are required where linking w/ OpenSSL libraries is required.

chipitsine:
OK, makes sense

chipitsine:
@johnmah , can you please add couple of builds to travis-ci build matrix ?

johnmah:
@chipitsine i added the static win32/64 builds to travis-ci. that work for you?

chipitsine:
@mattock , can you merge it please ?

chipitsine:
@johnmah , I'm not sure we need depcache + static builds

as for me, I would get rid of depcache

mattock:
This PR looks good afaics, so merging.

@chipitsine : I would keep depcache in there for now. I rarely use it, but when modifying openvpn.nsi it is quite useful as it cuts down compile time a lot.

PR #116, opened 2017-11-13
Actually accept a given value to --sign-digest in windows-nsis/build

gertvdijk:
Any value given to this arg will print the usage instead of being passed to
osslsigncode.

bug introduced in commit 0e3868e5

Signed-off-by: Gert van Dijk <gert@gertvandijk.net>


mattock:
This is trivial. Merging, thanks!

PR #117, opened 2017-11-19
AppVeyor: use proper way to resolve project path, also some cleanup

chipitsine:
APPVEYOR_PROJECT_SLUG point to project path properly, when
repo is cloned, deleted and cloned again


mattock:
All commits look reasonable. Merging.

PR #118, opened 2018-01-14
add openssl-1.1.0 support, also some cleanup

chipitsine:


.travis.yml
```diff
@@ -39,8 +38,13 @@ matrix:
       env: SCRIPT=windows-nsis/build-snapshot
     - compiler: ": complete"
       env: SCRIPT=windows-nsis/build-complete
+    - compiler: ": snapshot"
+      env: SCRIPT=windows-nsis/build-snapshot OPENSSL_VERSION=1.1.0g
   exclude:
     - compiler: gcc
+  allow_failures:
+    - compiler: ": snapshot"
+      env: SCRIPT=windows-nsis/build-snapshot OPENSSL_VERSION=1.1.0g
```
mattock: Is there a particular reason to repeat an identical "env:" line here?

chipitsine: it is how "allow_failures" works.

currently, even snapshot build fail on openssl-1.1.0g, so we (temporary) mark them as allowed failures.

mattock: Ok. Quick search [confirmed this](https://stackoverflow.com/questions/20905969/travisci-how-to-allow-failures-for-environment-variable#20909996).

generic/build
```diff
@@ -30,23 +30,20 @@
...
 empty_ifelse() {
-	local v="$1"
-	local t="$2"
-	local f="$3"
+	[ -z "$1" ] && echo "$2" || echo "$3"
+}
 
-	[ -z "${v}" ] && echo "$2" || echo "$3"
+contains() {
+    [ "${1#*$2}" != "$1" ] && echo "$3" || echo "$4"
```
mattock: This is an interesting construct and does seem to work as advertised. Do you happen to have a link where this is explained?

mattock: I mean the ```${1#*$2}``` part of course.

generic/build
```diff
@@ -30,23 +30,20 @@
 set -e
 
 die() {
-	local m="$1"
```
mattock: The removal of reassigning positional parameters into local-scope variables using "local" should be safe and makes the code a bit more readable. From Dash man-page:

_"When a shell function is executed, all of the shell positional parameters (except $0, which remains unchanged) are set to the arguments of the shell function.  The variables which are explicitly placed in the environment of the command (by placing assignments to them before the function name) are made local to the function and are set to the values given.  Then the command given in the function definition is executed.  The positional parameters are restored to their original values when the command completes."_

From Bash man-page:

_"The positional parameters are temporarily replaced when a shell function is executed (see FUNCTIONS below)."_

So there should be no risk of leaking the positional parameters from the function.

mattock:
Overall looks good so I'll merge this.

generic/build
chipitsine: somewhere in Bourne shell documentation, I cannot recall now.

I tried to keep it as much away from bash as possible

mattock: It seemed to work fine in Dash which tries to be POSIX-compliant. So this is definitely not specific to Bash.

chipitsine: I tested on /bin/sh under freebsd, which I suppose is Bourne shell

PR #122, opened 2018-01-24
make openvpn.nsi compatible with openssl-1.1.0 as well

chipitsine:


mattock:
@chipitsine can you modify the *.dll line to be more specific, preferably today? I will start building OpenSSL 1.1-based installers tomorrow evening so this PR would have to be merged by then.

Besides that one minor change this PR looks good.

chipitsine:
there're 2 dlls in each cases. I haven't found an elegant way to handle that.

chipitsine:
I think, we will stick to openssl-1.1.0 after a while and lock to those binaries.

mattock:
Then I think we should just list both OpenSSL 1.0.x and 1.1.x versions of the libraries in the uninstaller section for now. It's only a couple of extra lines.

chipitsine:
we can add files like

```
File /nonfatal "xxx1.dll"
File /nonfatal "xxx2.dll"
File /nonfatal "xxx3.dll"
File /nonfatal "xxx4.dll"
```

I'm not happy about those "nonfatal" switch.
making elegant solution to "include either 2 or other 2 files" is somewhat tricky within NSIS, I decided to stay with "*.dll"

mattock:
For the uninstaller section we should not need /nonfatal - we can have possibly non-existing files on that list. And for the SecOpenSSLDLLs your current approach should work and is probably the best we can do.

chipitsine:
uninstalling capabilities of NSIS suck. the best approach would be

1) track installed files
2) remove them

not trivial to implement using NSIS.

there might be side effects when lzo will be removed even if it was not installed

chipitsine:
I do not see any valuable benefits in uninstalling 4 files instead of "*.dll", actually

mattock:
Some users have surely installed .dll files into the OpenVPN bin/ directory. For example OpenVPN plugins. We should not wipe them out when doing normal uninstall/install cycle.

chipitsine:
makes sense. I'll change uninstall soon

mattock:
👍

PR #130, opened 2018-05-29
Update build.vars

TinCanTech:
Enable Windows only vars for Windows only builds.

### DIFF
--- a/generic/build.vars
+++ b/generic/build.vars
@@ -41,7 +41,7 @@ EXTRA_OPENSSL_CONFIG="${EXTRA_OPENSSL_CONFIG:--static-libgcc}" # uncomment if op
 #EXTRA_PKCS11_HELPER_CONFIG
 #EXTRA_OPENVPN_CONFIG
 
-EXTRA_TARGET_CFLAGS="-Wl,--dynamicbase,--nxcompat"
+[ $(echo $IMAGEROOT | grep -q 'win') ] && EXTRA_TARGET_CFLAGS="-Wl,--dynamicbase,--nxcompat"
 
 # You can enable --high-entropy-va for 64-bit builds if you have recent enough
 # mingw-w64


mattock:
This idea makes sense. Does this work so that if one is using windows-nsis/build-complete then the EXTRA_TARGET_CFLAGS get applied? What is the value of IMAGEROOT in that case?

TinCanTech:
Spanner in the works -- No this does not work for windows-nsis/build-complete.
In that case IMAGEROOT is set to /{pathto}/windows-nsis/tmp/image-i686

TinCanTech:
This works for `generic/build` and `windows-nsis/build-complete`:
```
[ $(echo $IMAGEROOT | grep 'win') ] && EXTRA_TARGET_CFLAGS="-Wl,--dynamicbase,--nxcompat"
[ $(echo $IMAGEROOT | grep 'i686') ] && EXTRA_TARGET_CFLAGS="-Wl,--dynamicbase,--nxcompat"
[ $(echo $IMAGEROOT | grep 'x86_64') ] && EXTRA_TARGET_CFLAGS="-Wl,--dynamicbase,--nxcompat"
```
Although, for some reason this does not work with `-q` ?

Any other tests I should try ?

selvanair:
Useful to make this option conditional, but  requiring IMAGEROOT to contain a word ("win") does not appear very robust. In our case the option being added is mingw specific, so what about parse CHOST for "mingw" ? And, avoiding another subshell for grep,

case "${CHOST}" in
     *mingw*) EXTRA_TARGET_CFLAGS="-Wl,--dynamicbase,--nxcompat";;
esac

TinCanTech:
Seems like a logical choice.  Quick tests show it works for build-complete and generic/build

Shall I submit again with @selvanair 's proposal ?

PR #131, opened 2018-05-30
Update build.vars

TinCanTech:
Enable Windows only vars for only Windows builds


selvanair:
This one looks good :)

chipitsine:
will it work on non bash ?

selvanair:
Any reason for suspecting that? Its only a simple pattern matching in case label -- nothing bash-specific.

TinCanTech:
I took a moment to run `build-complete` under `login:/bin/bash` and again `login:/bin/sh` on **Ubuntu 18.04** .. the results were 100% success. I am running the binary on Win10 as server right now :)

Issue #127, opened 2018-05-28
build-complete may delete /tmp


qdii:
https://github.com/OpenVPN/openvpn-build/blob/be29cd5c7575da438783bd11412c04f32dfde28c/windows-nsis/build-complete#L10

A [common value](https://en.wikipedia.org/wiki/TMPDIR) for TMPDIR is `/tmp`. If this variable is set, then build-complete will remove everything in it.

mattock:
Good catch. We should definitely document this and maybe even verify that TMPDIR is _not_ set to /tmp. Care to provide a PR?

chipitsine:
I recall erasing my home directory when pointing TMPDIR to it

chipitsine:
I guess the best approach would be

1. create tmp dir from scratch
2. stop build if TMPDIR already exists

mattock:
@chipitsine wouldn't that mean that user would have to wipe TMPDIR manually before every build?  

Perhaps, instead of wiping the TMPDIR entirely in build-complete, we could only wipe the build directories _inside_ it. So, even if TMPDIR was /home/<user> or /tmp there would be no/minimal damage.

chipitsine:
I did have a backup of home dir, so I did not worry much :)

PR #132, opened 2018-06-28
Fix build-complete logic

TinCanTech:
windows-nsis/build-complete --build-depcache exits gracefully on completion

### DIFF
--- a/windows-nsis/build-complete
+++ b/windows-nsis/build-complete
@@ -26,11 +26,9 @@ main() {
 				${WIN_SAVE_DEPCACHE:+--save-depcache=win-$arch} \
 				|| die "build ${arch}"
 
-		# Loop here if building the depcache:
-		[ -n "${DO_ONLY_DEPS}" ] && continue
-
 	done
 
+	if [ -z "${DO_ONLY_DEPS}" ]; then
 		ROOT="${TMPDIR}/installer" \
 			./build \
 			--installer-version="${INSTALLER_VERSION}" \
@@ -47,8 +45,7 @@ main() {
 			--sign-timestamp="${SIGN_TIMESTAMP_URL}" \
 			|| die "pack installer"
 
-	echo
-	if [ -z "${DO_ONLY_DEPS}" ]; then
+		echo
 		echo "NOTICE: OpenVPN Installers are at '${OUTPUT_DIR}'"
 	else
 		echo "NOTICE: depcache prepared for future use"
--- a/windows-nsis/build-complete
+++ b/windows-nsis/build-complete
@@ -7,18 +7,18 @@ die() {
 }
 
 main() {
-	rm -fr "${TMPDIR}"
+	rm -fr "${BUILD_TMPDIR}"
 
-	"${WGET}" ${WGET_OPTS} --directory-prefix=${TMPDIR} "${EASY_RSA_URL}" || die "get easy-rsa"
-	"${WGET}" ${WGET_OPTS} --directory-prefix=${TMPDIR} "${TAP_WINDOWS_INSTALLER_URL}" || die "get tap-windows"
-	"${WGET}" ${WGET_OPTS} --directory-prefix=${TMPDIR} "${OPENVPNSERV2_URL}" || die "get openvpnserv2"
+	"${WGET}" ${WGET_OPTS} --directory-prefix=${BUILD_TMPDIR} "${EASY_RSA_URL}" || die "get easy-rsa"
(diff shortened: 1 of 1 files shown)


TinCanTech:
To explain in more detail:

1. Remove the `# loop here ...` because it is a **for** loop and will otherwise goto next `arch` anyway. The `continue` is redundant, maybe there was a plan to put something else after it but it never materialised.

2. Move the `if [ -z "${DO_ONLY_DEPS}" ]` to before we kick off the Windows openvpn build because if we are building depcache as the --help says `	--build-depcache create depcache only (no program build)` satisfying that condition.

AFAICS that is all there is to it.  Now `--build-depcache` exits cleanly without the FATAL error (caused by `--openvpn-bin-tarball-i686=$(ls ${TMPDIR}/image-i686/openvpn-i686-*-bin.*) \` in the next section of `build-complete`).

Now run `--use-depcache` and the script builds openvpn with the new depcache.

TinCanTech:
Here is a wee test script you can use to check it (takes less than 20mins on my ub1604 VM):

```
#!/bin/sh

die ()
{
	timestamp "ERROR= $1"
	exit 1
}

timestamp ()
{
	for i in 1 2 3 4
	do
	echo "#"
	done
	echo

	[ -z "$1" ] || echo "  $1"
	echo "  $(date)"

	echo
	for i in 1 2 3 4
	do
	echo "#"
	done
}

delete_existing ()
{
	echo "WARNING: delete existing .."
	DELETEME="$(pwd)/tct-test"
	[ -d "$DELETEME" ] || die "delete existing: $DELETEME"
	rm -rf "$DELETEME"
	[ $? -eq 0 ] || die "Delete existing: $DELETEME failed"
	echo "has deleted $DELETEME"
}

START_dir="$(pwd)"

REDO="$1"
[ "$REDO" = "--redo" ] && delete_existing



MY_dir=$1
[ "$MY_dir" = "--redo" ] && MY_dir=""
[ -n "$MY_dir" ] || MY_dir="tct-test"
FULL_dir="$START_dir/$MY_dir"

[ -d "$FULL_dir" ] && die "Directory exists: $FULL_dir"

echo "Creating build in $FULL_dir"

FULL_stime=$(date +%s)
timestamp "START"
git clone https://github.com/TinCanTech/openvpn-build.git $MY_dir || die "git clone"

cd $MY_dir/windows-nsis || Die "cd $MY_dir"

BLDDC_stime=$(date +%s)
timestamp "BEGIN: ./build-complete --build-depcache"
./build-complete --build-depcache || die "build depcache"
#sleep 20
BLDDC_etime=$(date +%s)

USEDC_stime=$(date +%s)
timestamp "BEGIN: ./build-complete --use-depcache"
./build-complete --use-depcache || die "use depcache"
#sleep 20
USEDC_etime=$(date +%s)

timestamp "COMPLETE"
FULL_etime=$(date +%s)

BLDDC_tooks="$(((BLDDC_etime-BLDDC_stime)%60))"
USEDC_tooks="$(((USEDC_etime-USEDC_stime)%60))"
FULL_tooks="$(((FULL_etime-FULL_stime)%60))"
BLDDC_tookm="$(((BLDDC_etime-BLDDC_stime)/60))"
USEDC_tookm="$(((USEDC_etime-USEDC_stime)/60))"
FULL_tookm="$(((FULL_etime-FULL_stime)/60))"

echo "--build-depcache took $BLDDC_tookm mins $BLDDC_tooks secs"
echo "--use-depcache took $USEDC_tookm mins $USEDC_tooks secs"
echo "Total took $FULL_tookm mins $FULL_tooks secs"

```

selvanair:
TinCanTech wrote:
> To explain in more detail:
> 
> Remove the # loop here ... because it is a for loop and will otherwise goto next arch anyway. The continue is redundant, maybe there was a plan to put something else after it but it never materialised.

The "redundant continue" in the original looks like a left-over from the times we used to have separate installers for  32 bit and 64 bit. The extra indentation in the nsis build command moved out of the loop point to the same.

The patch looks good to me. That said, I don't do build-complete (just build in the generic folder), so never noticed this issue.

TinCanTech:
I have added a PR to fix build-complete $TMPDIR #127 but as my branch is ahead of master so git has "merged" the new PR into this one.  May be my git foo is not up to scratch. Awaiting feedback ..

TinCanTech:
I messed up the second patch

mattock:
@TinCanTech with Git you often need to rewind history before refactoring a PR. It always makes sense to keep "master" branch at the same level as upstream and create feature branches. Then, if you need to recreate a PR you just "git checkout master" and create a new feature branch from it. If you don't checkout master first, then your changes will get applied on top of your previous changes.

TinCanTech:
Thanks but the second patch I submitted completely undid the first.  So I resubmitted correctly because it seemed the cleanest way to progress.

PR #136, opened 2018-09-25
build: fix Windows MinGW build

lstipakov:
PKG_CHECK_MODULES macro is used to check the presence of given packages
and, if found, sets variables FOO_CFLAGS and FOO_LIBS, where FOO is a package
name and _CLAGS and _LIBS values are taken from package config, for example
from /lib/x86_64-linux-gnu/pkgconfig/liblz4.pc

Setting PKG_CONFIG to 'true' doesn't make much sense, since it makes
PKG_CHECK_MODULES to assume that any package is installed without any checks
and without setting FOO_CFLAGS/LIBS values.

This breaks, for example, Windows MinGW build on Windows build machine,
since it assumes the presence of liblz4 and enables callinf LZ4 API.
Because LZ4_LIBS is not set, we got linker errors.

Fix by removing PKG_CONFIG assignment to 'true'.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -30,7 +30,7 @@ for:
       only:
         - BUILDSYSTEM: cygwin
     build_script:
-      - "%CYGWIN%\\setup-%ARCH%.exe -g -q -P mingw64-i686-openssl,mingw64-x86_64-openssl,cygwin-devel,unzip,dos2unix,patch"
+      - "%CYGWIN%\\setup-%ARCH%.exe -g -q -P mingw64-i686-openssl,mingw64-x86_64-openssl,cygwin-devel,unzip,dos2unix,patch,liblz4-devel"
       - "%CYGWIN%\\bin\\bash -lc 'cp `which true` /usr/bin/man2html'"
       - "%CYGWIN%\\bin\\bash -lc 'cd /cygdrive/c/projects/%APPVEYOR_PROJECT_SLUG% ; generic/build'"
     artifacts:
--- a/generic/build
+++ b/generic/build
@@ -403,7 +403,6 @@ if [ -n "${BUILD_FOR_WINDOWS}" ]; then
 			-lgdi32 \
 		"
 	fi
-	export PKG_CONFIG="true"
 fi
 
 while [ -n "$1" ]; do


chipitsine:
we do build on App Veyor

how it turned out that we did not see that error there?

lstipakov:
> we do build on App Veyor
> 
> how it turned out that we did not see that error there?

Probably by luck. Do you have `liblz4` installed in Cygwin and if yes, which version?

In any case we double check if it supports `LZ4_compress_default()` and if yes, then you should run into this problem. If `liblz4` is not installed or doesn't support above method, then bug won't be triggered.

@mattock Can you confirm that we no **not** have `liblz4` installed in our AppVeyor? Here is build log from previous PR where bug presents:

https://ci.appveyor.com/project/mattock/openvpn-build/build/cygwin-73/job/q5kl87n0e4t30eyg

    checking for LZ4... yes
    checking for LZ4_compress_default in -llz4... no
    checking for LZ4_decompress_safe in -llz4... no
    		usable LZ4 library or header not found, using version in src/compat/compat-lz4.*
    
As you see, while we (wrongly) assume that LZ4 is installed, test for `LZ4_compress_default` fails so we do not use LZ4 and bug won't be triggered.

For comparison below is log from this PR where bug has been fixed:

https://ci.appveyor.com/project/mattock/openvpn-build/build/cygwin-74/job/k2h4wb29344il1n4

    checking for lz4.h... no
    		usable LZ4 library or header not found, using version in src/compat/compat-lz4.*
     
Here we detect that LZ4 is missing and do not try to use it.

mattock:
@lstipakov the cygwin setup command line in [.appveyor](https://github.com/OpenVPN/openvpn-build/blob/master/.appveyor.yml) is this:

    "%CYGWIN%\\setup-%ARCH%.exe -g -q -P mingw64-i686-openssl,mingw64-x86_64-openssl,cygwin-devel,unzip,dos2unix,patch"

So liblz4 is not explicitly installed. And it is probably not installed by default, as it is available as an option (see [Cygwin package search](https://cygwin.com/cgi-bin2/package-grep.cgi?grep=liblz4&arch=x86_64)).

chipitsine:
ok, so is it better to install lz4 on app veyor ?

lstipakov:
> ok, so is it better to install lz4 on app veyor ?

I don't think it matters from OpenVPN's perspective - if liblz4 is missing then OpenVPN will use own implementation, "compat-lz4".

chipitsine:
from the bug you are trying to fix ?

lstipakov:
Ah, if you want to reproduce it - then yes, you could install recent version of lz4 (1.8) and run generic/build (I think windows-nsis scripts also run it). You should get linker error since PKG_CHECK_MODULES won't write `-llz4` into LZ4_LIBS.

chipitsine:
can you add that to your PR please ?

lstipakov:
> can you add that to your PR please ?

Interesting, AppVeyor's build environment couldn't find lz4 when libliz4-devel is installed:

    [00:14:45] checking for LZ4... no
    [00:14:45] checking lz4.h usability... no
    [00:14:45] checking lz4.h presence... no
    [00:14:45] checking for lz4.h... no
    [00:14:45] 		usable LZ4 library or header not found, using version in src/compat/compat-lz4.*

On my machine both Cygwin and Msys2/MinGW work fine:

    checking lz4.h usability... yes
    checking lz4.h presence... yes
    checking for lz4.h... yes
    checking additionally if system LZ4 version >= 1.7.1... ok
    checking for LZ4_compress_default in -llz4... yes
    checking for LZ4_decompress_safe in -llz4... yes

chipitsine:
that's interesting

I'll have a look in few days

chipitsine:
I played a little bit.

I added ```cat ./generic/tmp/openvpn-2.4.6/config.log``` to .appveyor.yml

here's output:

```
configure:17031: checking for LZ4
configure:17038: $PKG_CONFIG --exists --print-errors "liblz4 >= 1.7.1 liblz4 < 100"
Package liblz4 was not found in the pkg-config search path.
Perhaps you should add the directory containing `liblz4.pc'
to the PKG_CONFIG_PATH environment variable
No package 'liblz4' found
Package liblz4 was not found in the pkg-config search path.
Perhaps you should add the directory containing `liblz4.pc'
to the PKG_CONFIG_PATH environment variable
No package 'liblz4' found
configure:17041: $? = 1
```

full log: https://ci.appveyor.com/project/chipitsine/openvpn-build/build/cygwin-204/job/0vnkd7ova67fx7o5

will have a look today

lstipakov:
@mattock can this be merged (it got one ack from @chipitsine)?

I mostly care about removing `export PKG_CONFIG="true"`, having it there essentially breaks pkgconfig functionality.

mattock:
@lstipakov sounds reasonable. I shall.

PR #139, opened 2018-10-31
Update build-env.bat

crazyevent:
support visual studio 2017 Enterprise


mattock:
Uses the enterprise version if community version does not exist, which is nice. Merging.

PR #140, opened 2018-11-01
Update wget.js

crazyevent:
wget.js add optional proxy support

### DIFF
--- a/msvc/wget.js
+++ b/msvc/wget.js
@@ -34,9 +34,13 @@ var ForWriting = 2;
 var adTypeBinary = 1;
 var adTypeText = 2;
 var adSaveCreateOverWrite = 2;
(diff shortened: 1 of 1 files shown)


mattock:
Based on cursory research into WinHttp.WinHttpRequest object this look correct, but I can't easily test this. How do you override the default proxy settings (i.e. no proxy)?

crazyevent:
It seems not the best way for proxy setting，any suggestions?

mattock:
Would configuring proxy settings outside of wget.js work? For example use a system-wide proxy.

crazyevent:
sounds good, I will try it.

flichtenheld:
File does not exist anymore in current version.

PR #142, opened 2018-11-29
Setup the registry key and directory for automatically started connections

selvanair:
This goes with  the change in openvpnserv2.exe starting service version
1.4.0. Automatic service start type now defaults to Automatic for fresh
installs. Else the previous state is preserved.

Requires OpenVPNServ2.exe version 1.4.0.0 to be available at
http://build.openvpn.net/downloads/releases/openvpnserv2-1.4.0.0.exe

Tested on Windows 7 using a local build of updated openvpnserv2 1.4.0.0


selvanair:
Do we want to continue using service 1.3.0.0 for future 2.4.x releases? That would make sense as the change involved here is somewhat major. If so, this PR may be rejected.

A downside of that would be degrading from 2.5 to 2.4.7+ will have the unexpected effect of OpenVPNService setup as auto and started. Moving back to 2.4.6 and earlier will always be a problem in either case.

VChernomYA:
[](https://github.com/users/VChernomYA/succession/invitation)

flichtenheld:
We have decided to remove the nsis/mingw buildsystem since it is not maintained anymore. Only the msi/msvc buildsystem will remain, unless someone steps up to maintain the other parts. This issue will be closed when the removal actually happens.

PR #144, opened 2019-01-23
Add support for building with OpenSSL 1.1.1a

schwabe:
OpenSSL 1.1.1a needs an explicit make command for the build to succeed.
Also reverse the test for the version and use the older install method
for OpenSSL 1.0.* method and the new one for everything else.


schwabe:
Tested with OPENSSL_VERSION=1.0.2l and OPENSSL_VERSION=1.1.1a

PR #146, opened 2019-02-20
Prepare for OpenVPN 2.4.7 (version change, openssl-1.1.0j patch)

mattock:
Building OpenSSL 1.1.0j with mingw-w64 fails unless we patch the
makefile on the fly. The patch will not apply cleanly to OpenSSL 1.1.1,
so documentation was added to make note of it, as well as for the
openvpn-build patch management system in general.

URL: https://github.com/openssl/openssl/issues/7653

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -35,3 +35,20 @@ Directories
 
     Please refer to the README files in the subdirectories
     for further information.
+
+Patches
+**************************************************
+
+    The "generic/patches" directory contains patches applied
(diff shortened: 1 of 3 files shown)


cron2:
ACK

mattock:
@TinCanTech the typo is harmless, but I did not like (being a perfectionist), so this fixes it: https://github.com/OpenVPN/openvpn-build/pull/148

PR #148, opened 2019-02-21
Fix harmless typo in the openssl 1.1.0j patch

mattock:


ikreb7:
Why is this PR still open?

mattock:
Hmm, looks like this got merged as part some other PR or by accident when pushing a release.

PR #149, opened 2019-03-09
travis-ci: switch to xenial image

chipitsine:


mattock:
Afaics this look reasonable.

PR #153, opened 2019-09-19
generic/build: use -O2 optimization

lstipakov:
It turns out that we haven't had any optimization
enabled for mingw Windows build, and, for example,
Visual Studio generated code was significantly faster.

This turns optimization on, and according to
performance tests, it increases download bandwidth
by up to 20% and now we're on par with Visual Studio.

Signed-off-by: lstipakov@gmail.com <lstipakov@gmail.com>

### DIFF
--- a/generic/build
+++ b/generic/build
@@ -252,10 +252,10 @@ build_dep() {
 }
 
 build_openvpn() {
-
 	echo "Build openvpn"
 	cd "${BUILDROOT}/openvpn-${OPENVPN_VERSION}" || die "cd openvpn"
 	autoreconf -ivf # this allows to specify github url as OPENVPN_URL
+	CFLAGS="${CFLAGS} ${OPT_OPENVPN_CFLAGS}"
 	./configure ${CONFIGOPTS} ${EXTRA_OPENVPN_CONFIG} \
 		$(empty_ifelse "${DO_STATIC}" --enable-pkcs11 --disable-plugins) \
 		--enable-lzo \
@@ -267,7 +267,6 @@ build_openvpn() {
 }
 
 build_openvpn_gui() {
-
 	echo "Build openvpn-gui"
 	cd "${BUILDROOT}/openvpn-gui"* || die "cd openvpn gui"
 	./configure ${CONFIGOPTS} ${EXTRA_OPENVPN_GUI_CONFIG} \
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -14,10 +14,14 @@ environment:
       ARCH: x86
       CYGWIN: C:\Cygwin
       CHOST: i686-w64-mingw32
+      RANLIB: gcc-ranlib
+      AR: gcc-ar
     - BUILDSYSTEM: cygwin
       ARCH: x86_64
       CYGWIN: C:\Cygwin64
       CHOST: x86_64-w64-mingw32
+      RANLIB: gcc-ranlib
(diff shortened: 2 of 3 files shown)


generic/build
```diff
@@ -252,10 +252,10 @@ build_dep() {
 }
 
 build_openvpn() {
-
 	echo "Build openvpn"
 	cd "${BUILDROOT}/openvpn-${OPENVPN_VERSION}" || die "cd openvpn"
 	autoreconf -ivf # this allows to specify github url as OPENVPN_URL
+	CFLAGS="${CFLAGS} -O2 -flto"
```
selvanair: The build scripts supposedly supports several platforms and compilers, so hard-coding -O2 here may not be the right approach. Though I've no idea whether anyone uses this for anything other than mingw builds.

Ideally -g -O2 should be added by `configure` which it will do if we undef CFLAGS. Not an option for Windows builds as we want to add --Wl,dynamicbase etc..

I would suggest to add "-O2 -flto" to EXTRA_TARGET_CFLAGS in build.vars or to TARGET_CFLAGS in this file (around line 350)  over-ridden by  --configuration specific values in the case statement around line 410.

lstipakov: > I would suggest to add "-O2 -flto" to EXTRA_TARGET_CFLAGS in build.vars or to TARGET_CFLAGS in this file (around line 350) over-ridden by --configuration specific values in the case statement around line 410.

This will modify `CFLAGS` value for all builds (openvpn and dependencies), including, for example openssl. This conflicts with openssl's own optimization flags, see both `O3` and `O2` below:

`i686-w64-mingw32-gcc  -I. -Icrypto/include -Iinclude -DOPENSSL_USE_APPLINK -DDSO_WIN32 -DNDEBUG -DOPENSSL_THREADS -DOPENSSL_NO_STATIC_ENGINE -DOPENSSL_PIC -DOPENSSL_BN_ASM_PART_WORDS -DOPENSSL_IA32_SSE2 -DOPENSSL_BN_ASM_MONT -DOPENSSL_BN_ASM_GF2m -DSHA1_ASM -DSHA256_ASM -DSHA512_ASM -DRC4_ASM -DMD5_ASM -DRMD160_ASM -DAES_ASM -DVPAES_ASM -DWHIRLPOOL_ASM -DGHASH_ASM -DECP_NISTZ256_ASM -DPADLOCK_ASM -DPOLY1305_ASM -DOPENSSLDIR="\"/etc/ssl\"" -DENGINESDIR="\"/lib/engines-1_1\"" -DL_ENDIAN -DWIN32_LEAN_AND_MEAN -DUNICODE -D_UNICODE -m32 -Wall -O3 -fomit-frame-pointer -D_MT  -O2 -static-libgcc -D_WINDLL -MMD -MF crypto/bf/bf_ofb64.d.tmp -MT crypto/bf/bf_ofb64.o -c -o crypto/bf/bf_ofb64.o crypto/bf/bf_ofb64.c`

selvanair: Then we may need something like OPENVPN_CFLAGS_OPT defined based on the compiler/platform and use it in build_openvpn(). Or something similar that doesn't break the build script on any of the supported platforms.

lstipakov: Added `OPENVPN_CFLAGS_OPT` define, which is set in `windows-nsis\build-complete` - it is only used to build windows installers.

cron2:
If selva says this is good, it is for me :-)

PR #159, opened 2019-11-08
Upgrade to Visual Studio 2019

rozmansi:
As openvpn's repo already switched MSVC to VS2019, the openvpn-build should follow.

### DIFF
--- a/msvc/build-env.bat
+++ b/msvc/build-env.bat
@@ -18,12 +18,12 @@ set OPENVPN_GIT=https://github.com/OpenVPN/openvpn.git
 set P7ZIP_URL=https://netcologne.dl.sourceforge.net/project/sevenzip/7-Zip/9.20/7za920.zip
 
 if "%ProgramFiles(x86)%"=="" set ProgramFiles(x86)=%ProgramFiles%
-if "%VSCOMNTOOLS%"=="" set VSCOMNTOOLS=%ProgramFiles(x86)%\Microsoft Visual Studio\2017\Professional\Common7\Tools
-if not exist "%VSCOMNTOOLS%" set VSCOMNTOOLS=%ProgramFiles(x86)%\Microsoft Visual Studio\2017\Community\Common7\Tools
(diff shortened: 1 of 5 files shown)


lstipakov:
@rozmansi could you cherry-pick this commit into your PR? https://github.com/lstipakov/openvpn-build/commit/d97a401bcb1e9afb180e3e073664f1e460db172d It installs strawberryperl and adds it in the beginning of PATH: https://ci.appveyor.com/project/lstipakov/openvpn-build/builds/28795052/job/verqim5tsn3edvww

msvc/build-env.bat
```diff
@@ -18,16 +18,18 @@ set OPENVPN_GIT=https://github.com/OpenVPN/openvpn.git
...
+if not exist "%VSCOMNTOOLS%" SET VCHOME=%ProgramFiles(x86)%\Microsoft Visual Studio\2019\BuildTools\VC
 
 if "%P7Z%"=="" SET P7Z=tools\7za.exe
 if "%GIT%"=="" SET GIT=c:\Program Files\Git\bin\git.exe
 if "%NASM_DIR%"=="" SET NASM_DIR=c:\Program Files\NASM
 
 if "%TARGET%"=="" SET TARGET=%ROOT%\image
 if "%RELEASE%"=="" SET RELEASE=Release
+
+if "%RC%"=="" SET RC=rc
```
lstipakov: Why this is needed?

rozmansi: My computer got RC defined as "1" somehow by OpenSSL build system. If %RC% is set before invoking OpenSSL build, it leaves it.

lstipakov: In fact I got the same problem on my environment, but wasn't able to trace origin. I solved it by running `set RC=`. Could you add a comment above that line with explanation (something like "on some environments OpenSSL defines RC as 1")?

rozmansi: Sure. Fixed and force-pushed (to upgrade existing commit, rather than introducing new).

lstipakov:
@mattock I think we can merge this

PR #160, opened 2019-11-09
windows-nsis: Refine the WoW64 decision logic

rozmansi:
The registry and file reflection is also present on ARM64 platform. Therefore, the reflection logic should resemble `${RunningX64} || ${RunningArm64}` or `${IsWow64}` or `!${RunningX86}` for short.


rozmansi:
This does not apply to OpenVPN until an ARM64 binaries are actually available.

PR #161, opened 2019-11-09
windows-nsis: Remove service dependency on tap0901

rozmansi:
When there are no TAP-Windows6 adapters present, the tap0901 driver's service is not (and cannot be) started. This prevents all OpenVPN services from starting. The no-TAP-Windows6/Wintun adapters present situation is checked at run-time.


selvanair:
Don't you mean "remove dependency on tap0901" ?

rozmansi:
> Don't you mean "remove dependency on tap0901" ?

Thanks. Updated the commit message and PR title.

lstipakov:
This is must-have for Wintun, where you could have only Wintun adapter installed by not tap-windows6.

PR #172, opened 2020-04-16
replace rfc7512 URI patch with latest version in Fedora

becm:
The `pkcs11-helper` patch for handling rfc7512 PKCS11 URIs has known flaws:
https://bugzilla.redhat.com/show_bug.cgi?id=1516474
https://bugzilla.redhat.com/show_bug.cgi?id=1825496

This got fixed in [Fedora](https://src.fedoraproject.org/rpms/pkcs11-helper/raw/master/f/pkcs11-helper-rfc7512.patch), which should have a large user base for `pkcs11-helper`.
All changes in patch are platform-independent so behaviour should be consistent.

Update might fix [#1075](https://community.openvpn.net/openvpn/ticket/1075) if problems are related to serial length.
Initial post not clear on this, the serial value looks "obfuscated".


becm:
[Patch orign](https://github.com/OpenSC/pkcs11-helper/pull/4) now includes [2nd bugfix](https://bugzilla.redhat.com/show_bug.cgi?id=1825496) as well.

mattock:
I've added this to today's meeting agenda:
* https://community.openvpn.net/openvpn/wiki/Topics-2020-04-23

mattock:
In yesterday's meeting we agreed that having a look at the Fedora's version of the patch would make sense as that is in relatively wide use and the code paths touched don't seem at all Windows-specific.

@dsommers is this the correct patch?

* https://koji.fedoraproject.org/koji/fileinfo?rpmID=20355303&filename=pkcs11-helper-rfc7512.patch

dsommers:
> @dsommers is this the correct patch?
> 
>     * https://koji.fedoraproject.org/koji/fileinfo?rpmID=20355303&filename=pkcs11-helper-rfc7512.patch

I would suggest to use what you find here: https://src.fedoraproject.org/rpms/pkcs11-helper/tree/master

This is the repository used by Koji to fetch the needed files doing the build.  Koji is just the build system for Fedora packages.  The master branch from the URL I'm pointing at is what is used in Fedora Rawhide (the "development" branch of the next Fedora).  Alternatively you can pull what is found in the latest 'f??' branch, which targets a specific Fedora release - but that means you need to check the branch each time you consider fetching a new patch.

I would suggest having a simple script downloading this patch, as when that patch disappears it means something has happened in the upstream pkcs11-helper.  And also if the patching fails, then we get a different notification that something has changed.

becm:
*Sorry in advance for the wall of text, but it seemed to me code origin and timeline were also a source of confusion in the 2020-04-23 meeting...*

Origin of work seems to be @dwmw2's pull request (*upstream*).

He integrated those changes in May 2015 Fedora [code](https://src.fedoraproject.org/rpms/pkcs11-helper/c/b6b420) to `pkcs11-helper 1.11`.

Release of `pkcs11-helper 1.22` was February 2017, also updated in Fedora.
In September 2017 copyright notice in openvpn-helper [changed](https://github.com/OpenSC/pkcs11-helper/commit/cb4d1131091ad1532713d57a491ed2ce8fbe03d4#diff-915980899741e320afbf1b178177f679), leading to single line offset between Fedora patch for `v1.22` and later *upstream* adaption.

@chipitsine added the patch (for v1.22 code) to openvpn-build in **October 2017**, possibly by stripping git headers from the Fedora patch at that time.

In **September 2017** a [bugfix](https://bugzilla.redhat.com/show_bug.cgi?id=1516474) was applied to [Fedora](https://src.fedoraproject.org/rpms/pkcs11-helper/c/133efa) for `pkcs11-helper 1.22` (old copyright notice) and [pull request](https://github.com/OpenSC/pkcs11-helper/pull/4#issuecomment-346803484) separately.
OpenVPN's `libpkcs11-helper-1.dll` no longer in sync with RH/Fedora due to outdated local patch.

In **April 2020** I was affected by `rfc7512` URI problems in OpenVPN and began to investigate. I identified the missing update and a **further problem** in the upstream code.

After verifying the Travis Windows/x64 build now works as expected (with my formerly affected hardware and CertID) using the above adjustments, I [submitted](https://bugzilla.redhat.com/show_bug.cgi?id=1825496) the additional problem and possible fix to the Fedora bug tracker.
This was [picked up](https://github.com/OpenSC/pkcs11-helper/pull/4#issuecomment-616624610) by @dwmw2 shortly after (force push 2020-04-20).

Beware when not applying the above changes directly:
- the *upstream* pull request is off by a line when applied to `pkcs11-helper v1.22`
- Fedoras [patch](https://src.fedoraproject.org/rpms/pkcs11-helper/blob/master/f/pkcs11-helper-rfc7512.patch) is currently **missing** the latest *upstream* bugfix (and has and git header referencing outdated commit IDs)

becm:
@dsommers @mattock beware when not applying the above changes directly:
- the *upstream* pull request is off by a line when applied to `pkcs11-helper v1.22`
- Fedoras [patch](https://src.fedoraproject.org/rpms/pkcs11-helper/blob/master/f/pkcs11-helper-rfc7512.patch) is currently **missing** the latest *upstream* bugfix (and has a git header referencing outdated commit IDs)

dwmw2:
Sorry, I got as far as updating https://github.com/OpenSC/pkcs11-helper/pull/4 but hadn't updated the Fedora packages. Should be done now.

becm:
To me it seems the build environment would have to be adapted to cleanly support downloaded patches, it already does some matching to find patches for extracted source packages. Most components to do more seem to be there but I am not confident to tinker around without creating a mess in the process.

As a intermediate step (and to match the result of the *2020-04-23* meeting) we could for now just replace the local copy with the [pristine Fedora patch](https://src.fedoraproject.org/rpms/pkcs11-helper/raw/master/f/pkcs11-helper-rfc7512.patch). Thanks to @dwmw2's latest update this should work without drawbacks.
*patch set update frequency of sample size 2 suggests; OK for the next 2-3 years*

If either updates `pkcs11-helper`, some sort of manual action will be needed anyways.

@dsommers any reservations to the *just make a static copy* approach until someone is in the mood to tackle build environment refactoring?

mattock:
I would be fine with just updating the patch we have in this repository.

dsommers:
For a quick solution, static copy is fine. But I think it way be worth looking into automating the patch fetching, and it seems like it will take quite some time before this issue will be resolved in pkcs11-helper natively.

That said, it is @mattock who has the final word.

mattock:
This is now part of https://github.com/OpenVPN/openvpn-build/pull/175

PR #175, opened 2020-07-03
MSI build automation prerequisites

mattock:
This PR adds support for code-signing executables and DLLs in "generic/build". This is required by MSI packaging. The code is essentially copy-and-paste from windows-nsis/build so it is well-tested. However, in the longer term - possibly not in the context of this PR - we might want to consider splitting off the signing functionality from both "generic/build" and "windows-nsis/build" and to do the signing as an extra step after the build.

The second part of this PR adds a brutal but functional Bash script for updating the last hexadecimal segment in PRODUCT_CODE with a new random value, as well as incrementing the package version in PRODUCT_VERSION. This script is primarily intended for use with buildbot or other CI systems, but can be used for release builds as well.


cron2:
I have not tried to understand each change in full detail, but from a slightly more distant point of view, all changes look reasonable, and I cannot see anything dangerous or obviously insecure here.  So, LGTM.

PR #185, opened 2020-08-26
Fix/no msi repair

rozmansi:


lstipakov:
Doesn't seem to work, unfortunately :(

I still can run `msiexec /fu` and also I see "repair/reinstall" radiobuttons.

rozmansi:
There's no way to prevent somebody from running `msiexec /fu`. MSI has no appropriate boundary to prevent that. This PR removes the "Change" or "Modify" button only.

rozmansi:
I see, PR/branch name was promising a different thing than it delivered. :( Added a patch to hide Repair button as well.

Mind that this is only hiding the broken MSI features from the Windows UI. Users still may play with msiexec command line.

lstipakov:
There are still "repair/remove" radiobuttons when you double-click on MSI file. Can we get rid of those, too?

rozmansi:
> There are still "repair/remove" radiobuttons when you double-click on MSI file. Can we get rid of those, too?

Ahh, totally forgot we have this nice UI for Modify/Repair/Remove too. I have removed the options from the GUI now.

For the record: I was very unhappy with the Wintun MSM merge module disabling the MSI rollback - I felt it back then that it would lead into all kind of anomalies. In the WireGuard MSI (no GUI, no change, no repair) it works. But, the OpenVPN MSI is a classic MSI package with GUI and everything.

In the long term, Wintun and TAP-Windows6 drivers will have to be installed differently. That's why I only commented out the Repair and Change MSI features for now.

lstipakov:
```
DEBUG: Error 2827:  The button &Remove on the radiobutton group Buttons on dialog MaintenanceTypePage extends beyond the boundaries of the group to the right by 80 pixels
The installer has encountered an unexpected error installing this package. This may indicate a problem with this package. The error code is 2827. The arguments are: MaintenanceTypePage, Buttons, &Remove
DEBUG: Error 2855:  The radiobutton group Buttons on dialog MaintenanceTypePage has fewer than 2 buttons
The installer has encountered an unexpected error installing this package. This may indicate a problem with this package. The error code is 2855. The arguments are: MaintenanceTypePage, Buttons, 
MSI (c) (44:08) [11:10:00:996]: Product: OpenVPN 2.5-20200304 amd64 -- The installer has encountered an unexpected error installing this package. This may indicate a problem with this package. The error code is 2855. The arguments are: MaintenanceTypePage, Buttons, 
```

rozmansi:
> 
> 
> ```
> DEBUG: Error 2827:  The button &Remove on the radiobutton group Buttons on dialog MaintenanceTypePage extends beyond the boundaries of the group to the right by 80 pixels
> The installer has encountered an unexpected error installing this package. This may indicate a problem with this package. The error code is 2827. The arguments are: MaintenanceTypePage, Buttons, &Remove
> DEBUG: Error 2855:  The radiobutton group Buttons on dialog MaintenanceTypePage has fewer than 2 buttons
> The installer has encountered an unexpected error installing this package. This may indicate a problem with this package. The error code is 2855. The arguments are: MaintenanceTypePage, Buttons, 
> MSI (c) (44:08) [11:10:00:996]: Product: OpenVPN 2.5-20200304 amd64 -- The installer has encountered an unexpected error installing this package. This may indicate a problem with this package. The error code is 2855. The arguments are: MaintenanceTypePage, Buttons, 
> ```

Unbelievable?! MSI has a check that there must be at least _two_ radio button items in a group. Wow, somebody at Microsoft really had too much time on his hands.

Rather than removing the Change and Repair radio buttons, the Continue button is disabled unless the Remove radio button is selected. I am aware this is not optimal UX, but as the TAP-Windows6 and Wintun setup will have to be revised to support Change and Repair properly, I plan to restore Change/Repair features later.

lstipakov:
Hm.. are commits up to date?

rozmansi:
> 
> 
> Hm.. are commits up to date?

They were not. This is me doing three different projects on three desktops at a time.

lstipakov:
LGTM. Tested that "Continue" is disabled unless "Remove" is chosen.

lstipakov (approved):

selvanair:
> There's no way to prevent somebody from running `msiexec /fu`. MSI has no appropriate boundary to prevent that. This PR removes the "Change" or "Modify" button only.

MSI is then a totally broken framework.  Using msiexec /fu any user can now remove the installed openvpn.exe or add utilities not installed by default by an admin.

PR #205, opened 2021-02-26
Bump OpenVPN version for snapshot build

becm:
Same issue as during [last version change](/OpenVPN/openvpn-build/commit/2eac27).

Was updated in OpenVPN repo [2020-08-12](/OpenVPN/openvpn/commit/f7432a9).
Error got masked due to previous problems with missing manpage (fixed by installing [python-docutils](/OpenVPN/openvpn-build/commit/ecc9771)).

### DIFF
--- a/windows-nsis/build-snapshot
+++ b/windows-nsis/build-snapshot
@@ -12,7 +12,7 @@ get_full_path() {
 }
 
 GIT="${GIT:-git}"
-OPENVPN_VERSION="${OPENVPN_VERSION:-2.5_git}"; export OPENVPN_VERSION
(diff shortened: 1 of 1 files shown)


cron2:
Like "yeah, obviously" :-)

becm:
A more permanent solution could be to execute
```sh
. version.sh
OPENVPN_VERSION="${OPENVPN_PACKAGE_VERSION}"
export OPENVPN_VERSION
```
after the configure step in `main()` (unconditional assignment, this is not really a choice here).

But this adds another complexity/dependency layer to the script.
On the other hand, it's already assumed implicitly (dist directory name)…

mattock:
@becm I'm all for a more permanent solution. Though in theory we are scrapping windows-nsis altogether, so a quick fix should be ok.

PR #209, opened 2021-03-22
Various msvc build fixes

rozmansi:
Thou 2.6 is targeting to replace MSVC with vcpkg & CMake, lets still keep things tidy, clean and working here.


lstipakov:
Started at the code and tested.

lstipakov (approved):

PR #210, opened 2021-03-23
Add LZ4 prerequisite building.

mattock:
commit 24596b258aa3a removed the bundled compat-lz4 from OpenVPN, thus
breaking windows/nsis builds with default config ("--enable-lz4").

Add download URLs and build invocation.  Since LZ4 has no "configure"
script, we pass the appropriate CC=/LD=/WINDRES= statement to "make".

Only static library is installed, since static linking is easier here
than to extend NSIS to bundle lz4.dll (+ whatever else is needed).

Tested with and without depcache.

32 and 64 bit binaries have been built, though only 64 bit binary was
tested (on Win10).

Signed-off-by: Gert Doering <gert@greenie.muc.de>


mattock:
I had to fix a few version issues in build.vars, otherwise this looks good:

```
cd openvpn-build/windows-nsis
./build-snapshot

--- snip ---

Output: "./openvpn-install-2.6_git-I601-Win10.exe"                                                                                                                         [42/1980]
Install: 7 pages (448 bytes), 20 sections (8 required) (20960 bytes), 2137 instructions (59836 bytes), 564 strings (40549 bytes), 1 language table (410 bytes).
Uninstall: 4 pages (256 bytes),                                                           
1 section (1048 bytes), 555 instructions (15540 bytes), 267 strings (4549 bytes), 1 language table (286 bytes).
Datablock optimizer saved 662170 bytes (~3.4%).                                           
                                                                                          
Using lzma (compress whole) compression.                                                  
                                                                                          
EXE header size:              100352 / 75776 bytes                                        
Install code:                          (122579 bytes)                     
Install data:                          (18264728 bytes)               
Uninstall code+data:                   (210603 bytes)                                     
Compressed data:             5226495 / 18597910 bytes
CRC (0xBD2D9FA3):                  4 / 4 bytes

Total size:                  5326851 / 18673690 bytes (28.5%)

NOTICE: OpenVPN Installers are at '.'
```

lstipakov (approved):
I can confirm that this change fixes the problem (I had to add --disable-lz4 before). I managed to successfully build NSIS installer, however not tested it.

PR #217, opened 2021-05-18
generic/build: build lz4 as dynamic library

lstipakov:
To be consistent with other dependencies and to avoid
cludge in build system to support both mingw and msvc
openvpn binaries, build lz4 as a dynamic library.

To make this change ARM64-proof, MSI part includes
customization of lz4 dll name. MinGW builds it as "liblz4.dll",
but MSVC (will be used for ARM64) uses "lz4.dll".

### DIFF
--- a/generic/build
+++ b/generic/build
@@ -239,19 +239,15 @@ build_dep() {
 	else
 		echo "Build lz4"
 		cd "${BUILDROOT}/lz4"* || die "cd lz4"
-		echo ./configure ${CONFIGOPTS} ${EXTRA_LZ4_CONFIG} \
-			$(empty_ifelse "${DO_STATIC}" --enable-shared --disable-shared) \
-			|| die "Configure lz4"
 		# lz4 does not have a configure, so pass CC/LD instead
-		# ... and we only want the static library
 		${MAKE} ${MAKEOPTS} CC=${CHOST}-gcc LD=${CHOST}-gcc \
 			WINDRES=${CHOST}-windres TARGET_OS=MINGW32 \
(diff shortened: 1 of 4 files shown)


cron2:
Hi,

On Tue, May 18, 2021 at 01:10:55AM -0700, Lev Stipakov wrote:
> To be consistent with other dependencies and to avoid
> cludge in build system to support both mingw and msvc
> openvpn binaries, build lz4 as a dynamic library.

I do not see a dynamic library as beneficial here.

Nothing else shares this library, and if we upgrade it, we'll never
upgrade the lz4.dll "as such" but "a new installer with all stuff".

So I would NAK this, unless someone can convince me that there is
anything beneficial about this.

(Packaging LZO as DLL is also not something I consider beneficial, but
that was decided before I came on board.  OpenSSL is, I think, shared
between GUI and openvpn.exe, and also openssl.exe, so there it makes
sense)

gert

-- 

lstipakov:
I could not find an easy way to mix static and dynamic libraries with vcpkg and MSBuild (which we currently use), although there seems to be a solution for CMake (https://github.com/microsoft/vcpkg/blob/master/docs/users/triplets.md#per-port-customization).

For MSBuild, there is project-wide setting "Use static libraries", and depends on it libraries are searched either in static or dynamic directory (vcpkg\installed\x64-windows or vcpkg\installed\x64-windows-static). One can manually add another directory on per architecture basis (like arm64-windows-static, x86-windows-static) but it looks like a hacky workaround to me.

lstipakov:
Found a way to link lz4 statically via custom triplets (https://patchwork.openvpn.net/patch/1815/) so this PR is not needed.

PR #218, opened 2021-05-20
ovpn-dco support

lstipakov:


lstipakov:
The resulting signed MSI installers could be found here:

- https://build.openvpn.net/downloads/snapshots/OpenVPN-2.6-dco-preview-amd64.msi
- https://build.openvpn.net/downloads/snapshots/OpenVPN-2.6-dco-preview-arm64.msi
- https://build.openvpn.net/downloads/snapshots/OpenVPN-2.6-dco-preview-x86.msi

TinCanTech:
I tested the x86 installer above on W7 and it all worked. Just one note, I was not prompted to stop the running GUI/Service as I was before.

lstipakov:
Please note that DCO driver is available starting from Windows 10 20H2. When running on earlier versions, "ovpn-dco" feature is hidden from installation dialog and driver is not installed.

It would be great if you could try the DCO on newer system.

TinCanTech:
x86 installs all components by default. Amd64 does not install Openvpn-Service or Easyrsa3 by default.

It would also be good if howto use DCO was documented in a sample file, I know it is opportunistic but having that in a sample file would be a good idea.  Also, I could find nothing in the log verb 4 about dco, so I don't know if it was in use.  The config is simple, using NCP and getting AES-256-GCM but I was using Peer-Fingerprint mode.

PR #220, opened 2021-06-16
MSVC: remove build scripts

lstipakov:
MSVC build, including dependencies, is natively supported
in openvpn repo, so build scripts here are not needed anymore.


cron2:
gone with the wind...

PR #221, opened 2021-06-17
generic/build: disable config autoload for OpenSSL

lstipakov:
On Windows cross-built OpenSSL loads config at path
c:\etc\ssl\openssl.conf, which is writable by non-admin user.

This could be a security risk, so disable config autoload.


cron2:
As discussed before, this sounds like the right approach for 2.4 / NSIS, for ease of implementation.

For 2.5 and MSI, we take the same route for now, until we can figure out a) whether openvpn config loading is really needed on windows (instead of using cryptoapi), and b) how to compile this in a secure way.

PR #230, opened 2021-11-18
COMMIT ON MASTER

jeremycorbeil86:


cron2:
Hi,

for a commit to be considered for inclusion, it at least needs a clear commit message that describes why the change is done, and what the effect is.  "COMMIT ON MASTER" is not a good commit message.

gert

PR #231, opened 2022-01-24
msi: switch to openssl3

lstipakov:


selvanair:
We also have to include `legacy.dll` -- this is what the user wil lhave to load using our new `--providers` option to use legacy algorithms. IIRC, based on the env vars we set, it would go to "install-path\modules\"

Also it would be nice to install the stock `openssl.cnf` in the distribution. For us the location would be "install-path\ssl\openssl.cnf". This one should not overwrite if already existing in the destination. What upstream build process does is to copy it as "openssl.cnf.new" and "openssl.cnf" with the latter skipped if it already exists.

lstipakov:
Legacy provider is now statically linked into libcrypto-3-x64.dll (https://github.com/lstipakov/openvpn/blob/ossl3/contrib/vcpkg-ports/openssl3/portfile.cmake#L38) and I smoke-tested it:

```
>	libcrypto-3-x64.dll!ossl_legacy_provider_init(const ossl_core_handle_st * handle, const ossl_dispatch_st * in, const ossl_dispatch_st * * out, void * * provctx) Line 198	C
 	libcrypto-3-x64.dll!provider_init(ossl_provider_st * prov) Line 905	C
 	libcrypto-3-x64.dll!provider_activate(ossl_provider_st * prov, int lock, int upcalls) Line 1103	C
 	libcrypto-3-x64.dll!ossl_provider_activate(ossl_provider_st * prov, int upcalls, int aschild) Line 1178	C
 	libcrypto-3-x64.dll!OSSL_PROVIDER_try_load(ossl_lib_ctx_st * libctx, const char * name, int retain_fallbacks) Line 31	C
 	libcrypto-3-x64.dll!OSSL_PROVIDER_load(ossl_lib_ctx_st * libctx, const char * name) Line 56	C
 	openvpn.exe!crypto_load_provider(const char * provider) Line 160	C
 	openvpn.exe!init_early(context * c) Line 120	C
 	openvpn.exe!openvpn_main(int argc, char * * argv) Line 228	C
 	openvpn.exe!wmain(int argc, wchar_t * * wargv) Line 376	C

```

lstipakov:
@mattock this change forces use of openss3, but we also have to build 2.5 releases with openssl1.1.1. How should we proceed? Make a 2.5 branch? Add parameters?

mattock:
I think a branch would be ok. I don't think we're going to make major changes to MSI code for release/2.5. So "master" branch in here could just track openssl 3.

cron2:
Hi,

On Wed, Feb 16, 2022 at 02:20:30AM -0800, Samuli Seppänen wrote:
> I think a branch would be ok. I don't think we're going to make major changes to MSI code for release/2.5. So "master" branch in here could just track openssl 3.

Yeah, this sounds like a good way forward.

master -> what is needed to build openvpn "master" (ossl 3)
2.5    -> what is needed to build openvpn 2.5.x (ossl 1.1.1*)

how to call that branch?  2.5, release/2.5, build/2.5?  I have no strong
opinion either way.

gert

-- 

lstipakov:
> I applied this patch and it works, except for one remaining reference to libcrypto_1_1. Here's a fix to it:

> -            <ComponentRef Id="bin.libcrypto_1_1.dll"/>
> +            <ComponentRef Id="bin.libcrypto_3.dll"/>

Looks like I screwed something when I was decoupling this from dco-related changes, since at least in one certain commit it looks correct: https://github.com/OpenVPN/openvpn-build/pull/232/commits/5ba5cbfce5171b148f4e5197454c958c04134ef2#diff-de1449325cad7b8ec578bc9ed7b5a53ece30e759ed64c8fa218e926c18dbec5bL1478

Anyway, I'll force-push the fix.

mattock:
Now it works. Let's have GitHub Actions finish, then merge this thing.

PR #232, opened 2022-01-27
Support for ovpn-dco

lstipakov:


mattock:
Looks good afaics. I will spin up a Windows buildbot worker on Monday and test this. We could move the DCO downloads to a more proper place, e.g. build.openvpn.net, though.

PR #242, opened 2022-04-22
bump pkcs11-helper to v1.29.0

becm:
Update to new `pkcs11-helper` release.
Adjust to archive path/type change (again) and incorporated win32 (cross-)compile fix.

The *old* `v1.28.0` archive was removed. So whithout this, builds fail for no (obvious) reason.


cron2:
looks reasonable

PR #245, opened 2022-05-10
windows-msi: adapt to manifest and presets changes in openvpn and ope…

lstipakov:
…nvpn-gui

With manifest, it is no longer needed to install dependencies separately, they're
installed as part of build process.

With presets, output directories are already set, no need to create build_<platform>.
Also CMAKE_TOOLCHAIN_FILE is set by presets from VCPKG_ROOT variable.


mattock:
Merging as discussed on the IRC.

PR #246, opened 2022-05-24
Update Easy-RSA to version 3.1.0

TinCanTech:


cron2:
NAK, as the patch should also update the easy-rsa version number in the line before the URL.

PR #247, opened 2022-05-24
Update Easy-RSA to version 3.1.0

TinCanTech:


cron2:
Have not tested, but if the master of EasyRSA says so, let's go there :-)

PR #248, opened 2022-05-25
Ossl1.1.1o

lstipakov:


mattock:
This seems to do the trick, provided  $Env:VCPKG_OVERLAY_PORTS is set correctly ("${basedir}\openvpn-build\windows-msi\vcpkg-ports").

PR #253, opened 2022-05-31
Add Powershell script for cleaning up build artefacts

mattock:
While debugging ninja issues in the OpenVPN 2.5.7 release cycle it became obvious that doing clean builds is hard, because lots of temporary files and artefacts are left lying around. The new script, cleanup.ps1, attempts to clean up all the build leftovers to allow doing truly clean builds.


mattock:
The script does its best to avoid nasty-looking errors, e.g. if directories it tries to clean up are not present (e.g. clean.ps1 ran already or no builds have not been made.

PR #255, opened 2022-06-02
update pkcs11-helper patch to match Fedora version

becm:
@kalev updated [pkcs11-helper](https://src.fedoraproject.org/rpms/pkcs11-helper/commits/master)  version and patch for Fedora.
Mirroring the patch update is to fulfill the ["do what Fedora does"](https://www.mail-archive.com/openvpn-devel@lists.sourceforge.net/msg19814.html) policy and have a *synchronized* state for current `pkcs11-helper`.

The patch used in Fedora is now (again) equivalent to taking commits directly from the respective [GitHub pull request](https://github.com/OpenSC/pkcs11-helper/pull/4.patch) changes.

Applying the new patch results in exactly the same changes to `pkcs11-helper` source code (minus the line offset warnings for versions after `v1.22` while doing so).

supersedes #203.

### DIFF
--- a/generic/patches/pkcs11-helper-001-RFC7512.patch
+++ b/generic/patches/pkcs11-helper-001-RFC7512.patch
@@ -1,38 +1,83 @@
-commit 90590b02085edc3830bdfe0942a46c4e7bf3f1ab (HEAD -> master)
-Author: David Woodhouse <David.Woodhouse@intel.com>
-Date:   Thu Apr 30 14:58:24 2015 +0100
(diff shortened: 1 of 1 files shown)


lstipakov (commented):
It is a bit hard to see what exactly has changed. Were there any code changes to patches or is it just reformatting?

becm:
Mainly pure reformatting.
Apart from the minor line offset adjustment it's identical (albeit segmented into the 3 original change sets) to the current patch (identical effect on source code, no warnings during application).

PR #259, opened 2022-06-22
GitHub Actions: use longer delay with port knocking

lstipakov:
Without delay or with small delay packets may arrive
in different order or blocked because they could be mistaken
for port scan. Use 500ms delay to try to avoid it.


mattock:
This makes perfect sense. Without this port knocking fails often, but not always. Hopefully with this it fails rarely.

PR #261, opened 2022-07-28
OpenVPNService: install by default and keep previous state

lstipakov:
As we don't support "Modify" action, adding OpenVPNService
is problematic since it requires uninstall and reinstall.

Install it by default. For clean install, use "Manual"
start mode and doesn't start it. For upgrade, preserve existing
configuration - for example if start mode was "automatic" and service
was started - the same will apply for the new installation.

This allows to remove code from openvpnmsica.dll which
reads service status, but doesn't apply it to the new installation.

### DIFF
--- a/windows-msi/msi.wxs
+++ b/windows-msi/msi.wxs
@@ -69,6 +69,11 @@
         -->
         <Property Id="REINSTALLMODE" Value="amus"/>
 
+        <Binary Id="Service.js" SourceFile="script\Service.js"/>
+
+        <CustomAction Id="CheckOpenVPNServiceStatus" BinaryKey="Service.js" JScriptCall="CheckOpenVPNServiceStatus" />
+        <CustomAction Id="ConfigureOpenVPNService"  BinaryKey="Service.js" JScriptCall="ConfigureOpenVPNService" Execute="deferred" Impersonate="no" />
+
         <!--
             Detect system information
         -->
@@ -78,9 +83,13 @@
         </InstallUISequence>
         <InstallExecuteSequence>
             <Custom Action="FindSystemInfo" After="FindRelatedProducts"/>
+            <Custom Action="CheckOpenVPNServiceStatus" After="ProcessComponents"/>
+            <Custom Action="ConfigureOpenVPNService" After="StartServices">NOT Installed</Custom>
         </InstallExecuteSequence>
         <UI>
             <ProgressText Action="FindSystemInfo">Detecting system information</ProgressText>
+            <ProgressText Action="CheckOpenVPNServiceStatus">Checking status of OpenVPNService</ProgressText>
+            <ProgressText Action="ConfigureOpenVPNService">Configuring OpenVPNService</ProgressText>
         </UI>
 
         <Property Id="TAPWINDOWS6ADAPTERS" Secure="yes"/>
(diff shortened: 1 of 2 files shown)


lstipakov:
Could be merged together with https://patchwork.openvpn.net/patch/2610/

selvanair:
This was in my wishlist but had no idea how to implement. Will test next week.
Unfortunately msi from the build artifacts do not work for upgrades as its version number is old. Is there a way to automatically update PACKAGE_VERSION (used to name the file) and PRODUCT_VERSION during CI builds?

lstipakov:
Yeah I think we can do it as part of GHa build. Meanwhile the easiest way would probably be to clone this, add a commit which updates variables you've mentioned and push. GHa should produce the installer.

lstipakov:
We also have a `bump-version.m4.ps1` and `bump-version.m4.sh` scripts which generate new product code (guid) and increment product version (like 2.5.10 -> 2.5.11). We could run those on CI build, but what's next? Do we need to commit those?

cron2:
I can merge the openvpnmsica patch, but "what is this javascript stuff doing" is beyond what I can quickly review... so if Selva can find time to test, that would be great.

On the "bump-version" scripts - is there a way to do some sort of persistant storage on GHA (so it knows on the next build which version that was...)?  Or we could just do timestamps (2.5.10.2134234)?

selvanair:
There is only 16 bits of space for the build number, timestamp may have to be very low resolution (1 per day?).  Also, I think we only need to change the number like 2.5.032 (major.minor.build) that goes into the MSI. Though its named "PRODUCT_VERSION" its not really the version of the release binaries. 

We could use GITHUB_RUN_NUMBER in place of "032" (build number) as it seems to be incremented for each run of a workflow?

selvanair:
This works as expected for (i) new installations where the automatic service gets installed and stopped, and (ii) for upgrades where the service was running. But for upgrades from a previous installation with automatic service not installed, this does not install the service at all. I expected to see the service ready for deployment, left in stopped/manual-start state.

lstipakov:
You mean at upgrade it is not installed by default? This is an expected behavior - MSI remembers the state of the component (not installed) which means you need to explicitly select it.

selvanair:
> You mean at upgrade it is not installed by default? This is an expected behavior - MSI remembers the state of the component (not installed) which means you need to explicitly select it.

So. it was my wrong expectation that we'll be able to change the defaults even in case of upgrades.

selvanair (approved):

lstipakov:
Since this is merged, @selvanair could you ack its counterpart in openvpnmsica? https://patchwork.openvpn.net/patch/2610/

selvanair:
This is merged only to master, not the release branch: so 2.5.8 installer doesn't install the service by default. Was that the intent?

TinCanTech:
I have tested W

selvanair:
Have you tested it with a fresh install? The purpose of this commit was to install the service by default on a fresh install. 

On upgrades it will always follow what the existing state is -- that is, will install and start the service if it was running or not install at all of it was not previously installed etc. The change is only for fresh installs where we want to change the default to always installed with startup set to manual, unless user explicitly selects another option.

TinCanTech:
I have tested a complete uninstall (v257), followed by a new install (v258).

selvanair:
The service binary is now always installed and left in the stopped state which is good. But, there is no way to have the service set to auto-start on a fresh installation. On upgrades everything works as expected as previous state of the service is respected, but after a fresh install one has to manually set the service to auto-start. Prior to this PR, selecting the custom service feature button would have  set it to auto-start, isn't it?  But it has lost that meaning now?

lstipakov:
> But it has lost that meaning now?

Yes. I've made a fix (improvement?) so that on clean installation service is auto-started.

https://github.com/OpenVPN/openvpn-build/pull/308

PR #264, opened 2022-08-19
Dco final

lstipakov:


lstipakov:
This is the same as another pull request (https://github.com/OpenVPN/openvpn-build/pull/232) but without commit which points openvpn branch to dco.

mattock:
I cannot spot any obvious problems with this, but I'm not fluent with WiX. Obviously @rozmansi would be a better person to review the actual code changes.

rozmansi:
I have reviewed this PR. It looks good to me. I have also compiled and packaged MSI and tested it to confirm it installs and uninstalls as advertised.

One suggestion (optional to implement) would be to use conditionalized components rather than features. Features are parts of the program a user selects based on his/hers preference. Well, computer may suggest initial feature selection and/or hide some.
When upgrading, the selected feature list it migrated to the new MSI. However, if the MSI is reinstalled as the part of the Windows 10-->11 upgrade, I see a potential problem here, as the `Drivers.OvpnDco.Nx20` feature is demoted and `Drivers.OvpnDco.Nx21` feature promoted. Personally, I would make all .inf/sys/cat components a part of the same universal `Drivers.OvpnDco` feature while introducing a `Condition` attribute to specific `<Component>` elements.
In other words, feature presence or absence defines whether a user want to install the DCO driver or not in general. It's something a human decides. While authoring the decision logic which files are correct using the component conditions. Which is something computer decides.

lstipakov:
As @rozmansi suggested, I switched to the single "Data Channel Offload" feature and made Win10/11-specific components conditional. 

Here are MSI packages built with this openvpn-build branch and an additional commit with points to the latest dco-win code in openvpn:

https://github.com/lstipakov/openvpn-build/actions/runs/2902240643

TinCanTech:
Tested `x86` on Windows-7, all OK.

Uninstalled old / Installed / Uninstalled new / installed, without any problems.

Note: The count-down timer for **uninstall** counts up not down.

lstipakov:
@mattock @rozmansi if this looks good, feel free to approve!

lstipakov:
No I think we're good.

PR #266, opened 2022-09-09
No need for running util/mkdef.pl

miiip:
1. I've changed the OpenSSL version from build.vars from 1.1.1k to 3.0.5
2. I've changed the OpenVPN version from build.vars from 2.5.4 to 2.5.7
3. I've removed the call to the Perl script utils/mkdef.pl since it is no longer needed when building OpenSSL 3.0.5.


flichtenheld:
We have decided to remove the nsis/mingw buildsystem since it is not maintained anymore. Only the msi/msvc buildsystem will remain, unless someone steps up to maintain the other parts. This issue will be closed when the removal actually happens.

flichtenheld:
OpenVPN master now switched to a new CMake+VCPKG buildsystem for Windows builds. That will be backported to release/2.6. No further support of the old "generic" build is planned.

PR #267, opened 2022-09-20
GitHub Actions: automatic versioning

lstipakov:
To enable upgrade of MSI snapshots, produce
installers with unique product code and incremental product version (2.6.x), where x is per-workflow
automatically incremented GITHUB_RUN_NUMBER.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -62,6 +62,17 @@ jobs:
           Invoke-WebRequest -Uri "https://wixdl.blob.core.windows.net/releases/v3.14.0.5722/wix314-binaries.zip" -OutFile wix.zip
           Expand-Archive -Path .\wix.zip -DestinationPath wix\bin
 
+      - name: Bump version
+        working-directory: openvpn-build/windows-msi
+        run: |
+          $NewProductCode = (New-Guid).ToString().ToUpper()
+          $BuildVersion = 10000 + [int]$env:GITHUB_RUN_NUMBER
+          $NewProductVersion = "2.6.$BuildVersion"
+          echo $NewProductCode $NewProductVersion
+          $version_m4 = (Get-Content version.m4)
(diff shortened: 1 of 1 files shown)


lstipakov:
Here are two consecutive builds with incremented version (2.6.142 and 2.6.143) and unique product code (required for upgrade):

 - https://github.com/lstipakov/openvpn-build/actions/runs/3090021241
 - https://github.com/lstipakov/openvpn-build/actions/runs/3090224143

lstipakov:
As discussed with @cron2 on IRC, use 1000 + GITHUB_RUN_NUMBER as build number.

selvanair:
> As discussed with @cron2 on IRC, use 1000 + GITHUB_RUN_NUMBER as build number.

This is definitely better and may keep CI builds always higher in version than releases. Installing releases on top of a CI build would require an uninstall, but that's manageable.

That said, I believe Simon's intent was to use `100*release-version + build` as the MSI version of release builds. See comment in windows-msi/version.m4 -- looks like we did not follow it for 2.5 as the last version was something like 2.5.036 (note the leading zero).

Anyway, if a future 2.6.12 release uses 2.6.1200 in the msi,  1000+ may not be just enough. Add 5000 instead? -- that still leaves enough room in the 16 bit integer.

selvanair:
Also, I'm assuming that we are only changing the MSI version. The product version in binaries (openvpn.exe etc.) will remain whatever is set there.

cron2:
Hi,

On Tue, Sep 20, 2022 at 08:54:56AM -0700, Selva Nair wrote:
> Anyway, if a future 2.6.12 release uses 2.6.1200 in the msi,  1000+ may not be just enough. Add 5000 instead? -- that still leaves enough room in the 16 bit integer.

Good argument.  This is why I wanted to wait for your feedback ;-)

We can add 10000, which would make nice numbers ("if it's 5 digits,
it's a private build"), and there's still room for 50.000 builds ;-)

gert
-- 

selvanair (approved):

PR #271, opened 2022-10-28
improve documentation for version bump

flichtenheld:


windows-msi/version.m4
```diff
@@ -40,8 +40,6 @@ dnl The package version as displayed by UI and used in filenames (no spaces, ple
 define([PACKAGE_VERSION], [2.6git])
 
 dnl The MSI product version in the form of n[.n[.n]] (numbers only).
-dnl The third field is 100*product release + package version.
-dnl The fourth field is ignored by MSI.
```
selvanair, line 44: 100*release + build was the original intent, but somehow we failed to follow it in practice.  In 2.6 we may want to revert to this so that the MSI version can be easily related to the openvpn version?
Also see: https://github.com/OpenVPN/openvpn-build/pull/267#issuecomment-1252561709

flichtenheld: Fair enough. I have instead replaced them with a better explanation of the original intent.

lstipakov (approved):

PR #273, opened 2022-10-31
Bump OpenSSL to 1.1.1s

lstipakov:


lstipakov:
I have not followed 2.5.8 release process (been on vacation), but if those are already build and we need 1.1.1q then we need to rebuild those.

From what I see https://www.openssl.org/news/openssl-1.1.1-notes.html there are no high/critical vulnerability fixes in 1.1.1q/p  . So if it is too much work to rebuild installers we could update OpenSSL in 2.5.9.

PR #275, opened 2022-11-01
windows-msi/version.m4: Update Easy-RSA to v3.1.1

TinCanTech:
Update Easy-RSA download source: OpenVPN/easy-rsa Release v3.1.1

Add new comment to clarify:
The 64bit version of Easy-RSA is suitable for
both 32bit and 64bit OpenVPN-Windows-MSI-Installers.

Note:
OpenSSL binaries from Easy-RSA zip file are not used by Openvpn-build. Openvpn-build uses OpenSSL binaries from vcpkg port.


lstipakov (approved):
Looks good - tested installers and checked that EasyRSA 3.1.1 is indeed used.

PR #279, opened 2022-11-15
Prevent concurrent run of t_client tests

lstipakov:
Test backend doesn't allow multiple connections with the same CN, so prevent running t_client aws tests in parallel.


cron2:
LGTM, matches documentation in https://docs.github.com/en/actions/using-jobs/using-concurrency

cron2 (approved):

PR #282, opened 2022-11-23
Bump tap-windows6 driver to 9.24.8

lstipakov:


flichtenheld:
I think I would prefer to merge this via release/2.5 to avoid having every commit duplicated going forward

lstipakov:
> I think I would prefer to merge this via release/2.5 to avoid having every commit duplicated going forward

https://github.com/OpenVPN/openvpn-build/pull/284/files ?

flichtenheld:
Superseded by #284

PR #286, opened 2022-11-25
GitHub Actions: don't upload t_client logs to remote

lstipakov:


flichtenheld (approved):
Makes sense

PR #293, opened 2022-12-05
Builde legacy.dll into MSI

lstipakov:
Install it to

    <installpath>\ssl\modules

which is set as value for OPENSSL_MODULES so that
openssl will be able to load legacy providers from there.


cron2:
Change looks good to me.  So as soon as someone has tested one of the resulting installers, ACK.

cron2:
Hi,

On Mon, Dec 05, 2022 at 09:11:32AM -0800, Frank Lichtenheld wrote:
> Merged #293 into release/2.6.

Not fully understanding the flow of "openvpn-build" - but is this also
merged to master?  If not, I think it should :-)

gert
-- 

PR #299, opened 2022-12-13
Write actual install path into openvpn-plap-install.reg

lstipakov:
openvpn-plap-install.reg has install path hardcoded, which breaks when OpenVPN is installed to non-default location.

Add 2 custom action:

 - GetInstallDir to get install path

 - UpdatePlapReg to update openvpn-plap-install.reg

 Second action is deferred and runs under elevated privileges,
 so it is able to modify files at installation path.


lstipakov:
@selvanair How does it look like?

flichtenheld:
I don't feel qualified to approve this. Would certainly prefer for @selvanair to judge.

selvanair (approved):
Looks good and works as expected.
The regexp \w: to match "C:" is probably a bit too lax, but looks okay for this purpose.

PR #303, opened 2022-12-16
bump-version.m4.sh: Fix to generate actual UUIDs

flichtenheld:

### DIFF
--- a/windows-msi/bump-version.m4.sh
+++ b/windows-msi/bump-version.m4.sh
@@ -10,11 +10,8 @@ PACKAGE_VERSION=`echo $PRODUCT_FULL_VERSION|cut -d "." -f 3|sed s/^0*//g`
 
 # Get current product code
 PRODUCT_CODE=`grep -E 'define\(\[PRODUCT_CODE' version.m4|cut -d " " -f 2|tr -d '[{}])'`
-PRODUCT_CODE_FIRST_SECTION=`echo $PRODUCT_CODE|cut -d "-" -f 1-4`
-
 # Create new product code
-PRODUCT_CODE_LAST_SECTION_NEW=`openssl rand -hex 8|tr -s '[:lower:]' '[:upper:]'`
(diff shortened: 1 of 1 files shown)


lstipakov:
Interesting, how has this been working before?

```
diff --git a/windows-msi/version.m4 b/windows-msi/version.m4
index c21e983..12bd505 100644
--- a/windows-msi/version.m4
+++ b/windows-msi/version.m4
@@ -46,10 +46,10 @@ define([PACKAGE_VERSION], [2.7git])
 dnl The MSI product version in the form of n[.n[.n]] (numbers only).
 dnl The third field is 100*openvpn bugfix release + MSI build number.
 dnl So for the 2nd MSI build for OpenVPN 2.6.3 use 2.6.302
-define([PRODUCT_VERSION], [2.7.0])
+define([PRODUCT_VERSION], [2.7.001])

 dnl The MSI product code MUST change on each product release.
-define([PRODUCT_CODE], [{9122FCB2-2BB5-4115-8C0F-05D9B5EFE8D6}])
+define([PRODUCT_CODE], [{9122FCB2-2BB5-4115-8C0F-4416BF7000710B9D}])

 dnl The MSI upgrade codes MUST persist for all versions of the same product line.
 dnl Please use own upgrade codes when deploying a non-official OpenVPN release.
```

The new one is indeed not valid GUID.

While this change makes sense, do we need this file at all? Since we are removing generic buildsystem and build on Windows,
shouldn't `bump-version.m4.ps1` be enough?

flichtenheld:
> While this change makes sense, do we need this file at all? Since we are removing generic buildsystem and build on Windows, shouldn't `bump-version.m4.ps1` be enough?

I still find this useful, since this change needs to be committed and pushed in the case of the release builds, which I do not want to do from the Windows build agent, but rather beforehand.

lstipakov (approved):

PR #308, opened 2022-12-23
Start service by default

lstipakov:
Instead of using "manual" start type,
start service by default for clean installs.


lstipakov:
ping @selvanair

selvanair (approved):
This is a much simpler approach than I had in mind. So, we'll always start the service on fresh install (unless the user deselects the service feature).

For truly fresh installs this is perfectly fine as config-auto will be empty. Those who are doing a clean re-install with some configs in config-auto folder, but had automatic service disabled may find this mildly surprising. That said, I like this approach.

Lightly tested in fresh install and update scenarios.

PR #310, opened 2022-12-29
debian-sbuild: port freight-add-many.py to Python 3

flichtenheld:
This was done some time ago, but seems we never
committed it.


mattock (approved):
Look good, thanks!

PR #311, opened 2022-12-29
Fix line continuation issue in sign-openvpn.bat

becm:
The x86 DLLs of `openvpn-gui` in `v2.6_rc1` did NOT get signed due to this.
Arguments to `signtool` stopped after `openvpn-gui.exe`.

Not sure what happens when the script tries to execute the (PLAP?) DLL (next command) , but it's definitely not what has been intended.


lstipakov (approved):
Indeed, for multiline commands in batch file we use caret (`^`) at the end of previous line and continue on the next line.

PR #316, opened 2023-01-10
Fix config files migration

lstipakov:
Starting from 2.5beta3

  https://github.com/OpenVPN/openvpn-build/commit/16a56571e4ee9a980496b3f19d8774d43ccfdea7

OpenVPN service uses separate directory (config-auto) for connections which require auto-start. Because of that, a migration functionality was added which moves config files from config to config-auto directory on upgrade if config-auto is missing and service is running.

Starting from 2.6rc1

  https://github.com/OpenVPN/openvpn-build/commit/1d15f8e4c5616d17fcfe744f7f9637b8f8263be7

we always install OpenVPNService, since it is problematic to add it due to lack of "Modify" action. Because of that change, when upgrading from pre 2.5beta3 to 2.6rc1 and newer, config files are always moved from config to config-auto. This happens because condition "should we start service" has been moved from Wix script to JS custom action, but the "upgrade" component in Wix script was dependent on that condition.

This commit fixes config migration functionality by moving it to JS custom action which handles OpenVPNService installation. Migration is only peformed when:

 - this is an upgrade and not clean install
 - service needs to be started
 - config directory exists
 - config-auto directory is missing in registry and on disk

As a side effect, we've lost "Where_are_my_config_files.txt" copying to config directory on migration. This can be fixed later.

Reported-by: Gert Doering <gert@greenie.muc.de>

### DIFF
--- a/windows-msi/msi.wxs
+++ b/windows-msi/msi.wxs
@@ -829,17 +829,6 @@
                         <Component Id="config.README.txt" Guid="{9C3B43AF-1399-4E6A-92CA-D7BA3B6BCC3C}">
                             <File Id="config.README.txt" Name="README.txt" Source="!(bindpath.build)README-config.txt"/>
                         </Component>
-                        <!-- Move auto-start config files from configuration folder to auto-start configuration folder on update from non-auto-start aware installations. -->
-                        <Component Id="config.Where_are_my_config_files.txt" Guid="{2DFEDF05-F41E-4EB0-A90A-DC76A261E444}">
-                            <Condition><![CDATA[CONFIGDIRREG AND NOT CONFIGAUTODIRREG]]></Condition>
-                            <File Name="Where are my config files.txt" Source="!(bindpath.build)Where are my config files.txt"/>
-                            <CopyFile
-                                Id="config_auto"
-                                SourceProperty="CONFIGDIRREG"
-                                SourceName="*"
-                                DestinationProperty="CONFIGAUTODIR"
-                                Delete="yes"/>
-                        </Component>
                     </Directory>
 
                     <Directory Id="CONFIGAUTODIR" Name="config-auto">
@@ -1678,7 +1667,6 @@
                 <ComponentRef Id="bin.libopenvpn_plap.dll"/>
                 <ComponentRef Id="bin.openvpn_plap_install.reg"/>
                 <ComponentRef Id="bin.openvpn_plap_uninstall.reg"/>
-                <ComponentRef Id="config.Where_are_my_config_files.txt"/>
                 <ComponentRef Id="config_auto.README.txt"/>
                 <ComponentRef Id="reg.autostart_config_dir"/>
 
--- a/windows-msi/script/Service.js
+++ b/windows-msi/script/Service.js
@@ -25,6 +25,33 @@ function IsServiceExist() {
     return svcs.Count > 0;
 }
 
+/*
+ * Checks if config migrations is required and if yes,
+ * returns [src, dst] directories.
+ *
+ * Config migration is required if:
+ *  - config_dir exists in registry
+ *  - autostart_config_dir doesn't exist in registry
+ *  - autostart_config_dir doesn't exists on disk
+ *
+ * Migration is performed only when service exists and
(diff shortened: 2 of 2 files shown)


lstipakov:
I also tested by:

 - installing 2.4.12 with service
 - starting service
 - copying ovpn profile to config
 - installing fixed MSI
 
As expected, profile was copied from config to config-auto.

selvanair (approved):
Changes look good to me.

Tested the following cases:
(i) clean install with config and config-auto directories present with user assets (ovpn's etc.) but no registry keys
(ii) upgrade from 2.6_rc1
(iii) upgrade from messy 2.4.12 (NSIS) with config-auto directory present but no reg-key for it and no service running
(as would happen if user did a downgrade from 2.5 to 2.4).
(iv) upgrade from clean 2.4.12 (NSIS) with service running

All worked as expected.

(v) same as (iii) but service running
As expected, in this case ovpn's in config folder are not copied to auto-config folder as the latter already exists. This does cause a surprising behaviour: ovpn's in config are autostarted before,  those in config-auto (if any) are autostarted after. I do not think we can do much about this without causing loss of files in config-auto folder.

selvanair:
Where is the source for README.txt added to config and config-auto folders? Would like to edit the latter to indicate that the GUI does scan this folder and list any config files found as persistent connections. Could also add a line that if PLAP is enabled, these connections are accessible from the login screen as well.

TinCanTech:
Would it be possible to add a dialogue box to explain the changes, after the install completes ?

selvanair:
In the old NSIS installer, we have a "Show README" at the end which is selected by default. I'm sure almost everyone hates that popup --  I do. Who wants unsolicited information. Post-installation dialogs may evoke a similar feeling. That said, we could display some text in the progress window for critical things like config files moved so that its there for those who read. No popup dialogs please.

As for new features, better for users to search and find out -- people are seldom receptive of good information when they are not seeking it. What about a menu in the GUI to list what's new? Also, things like context help in the GUI, good documentation that would show up in search etc. would be helpful.

TinCanTech:
For the specific case of moving the config files, I would prefer a dialogue box, no matter how annoying.

OpenVPN does not need to care about such insignificant trivialities and this would only show in the event that files have been moved **by the installer**.

The README which was ALWAYS presented before was over the top, this is one specific issue and only displayed when files have moved.  The two are not a valid comparison.

selvanair:
> dialogue box to explain the changes

I misread that to mean "explain what is new in this version". Sure, while moving files, we must inform the user. We do add "Where_are_my_config_files.txt" to the config folder from where files are moved, so those who who notice their files have moved will also see the explanation in there. We could also show a popup in addition to that.

However, as @lev mentioned

`As a side effect, we've lost "Where_are_my_config_files.txt" copying to config directory on migration. This can be fixed later.`

We have to remember to fix this.

selvanair:
> FTR: This is also a change in the installer behavior.

No its not a change. Starting 2.5 we alsway move config to config-auo if some conditions are satisfied. We recently broke it in some corner cases and this PR is to fix that. Nothing new. No need to ask for confirmation etc., as moving is the correct behaviour. Documenting/informing why the move is needed is all that may be lacking.

TinCanTech:
> No its not a change. Starting 2.5 we alsway move config to config-auo if some conditions are satisfied

Ok, I was not aware of this in 2.5, anyway, just feedback, sorry for the noise.

TinCanTech:
I only found out when @cron2 did.

Still, this has helped a little `;-)`

lstipakov:
> `As a side effect, we've lost "Where_are_my_config_files.txt" copying to config directory on migration. This can be fixed later.`
> 
> We have to remember to fix this.

See https://github.com/OpenVPN/openvpn-build/pull/317

lstipakov:
> Where is the source for README.txt added to config and config-auto folders? 

https://github.com/OpenVPN/openvpn-build/blob/master/windows-msi/doc/config/README.txt.in
https://github.com/OpenVPN/openvpn-build/blob/master/windows-msi/doc/config-auto/README.txt.in

Issue #214, opened 2021-05-10
Launch on user logon setting lost on upgrade


richardschuetz:
OpenVPN GUI is not launched on Windows startup anymore after upgrading the previous version to OpenVPN 2.5.2. The relevant Active Setup configuration is not working as intended:

```
HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Active Setup\Installed Components\{E5931AF4-2A8F-48A5-AFC8-0E8A268358A0}
    (Default)    REG_SZ    OpenVPN 2.5.2-I601 amd64
    Version    REG_SZ    1
    IsInstalled    REG_DWORD    0x1
    DontAsk    REG_DWORD    0x2
    StubPath    REG_EXPAND_SZ    reg add HKCU\Software\Microsoft\Windows\CurrentVersion\Run /f /v OPENVPN-GUI /t REG_SZ /d "C:\Program Files\OpenVPN\bin\openvpn-gui.exe"

HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Active Setup\Installed Components\{E5931AF4-2A8F-48A5-AFC8-CE9B79C4B19D}
    (Default)    REG_SZ    OpenVPN 2.5.1-I601 amd64
    Version    REG_SZ    1
    IsInstalled    REG_DWORD    0x0
    DontAsk    REG_DWORD    0x2
    StubPath    REG_SZ    reg delete HKCU\Software\Microsoft\Windows\CurrentVersion\Run /v OPENVPN-GUI /f
```

`{E5931AF4-2A8F-48A5-AFC8-CE9B79C4B19D}` is processed after `{E5931AF4-2A8F-48A5-AFC8-0E8A268358A0}` and immediately deletes the Run registry key again. It looks like the keys are processed in alphanumeric order, so only specific GUID combinations trigger this behavior.

kroimon:
This seems to be still an issue - we ran into the same problem.

With the NSIS installer, this never happened because it only ever wrote to `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Active Setup\Installed Components\OpenVPN_UserSetup`, but the MSI setup uses a different key with each version.

According to https://helgeklein.com/blog/active-setup-explained/, the `reg delete` command should never run because `IsInstalled` gets set to 0, but it looks like this is not true (anymore).

[...]

PR #320, opened 2023-01-12
ActiveSetup: fix "Run on logon" behavior

lstipakov:
Starting from 2.5, "Run on logon" installer feature didn't work in cases when ProductCode of new version is "less" than ProductCode in old version in in alphanumeric comparison.

 - Install version with product code 456
 - Log off and log on so that ActiveSetup command will be added to HKCU
 - Install version with product code 123
 - Log off and log on

Here openvpn-gui won't be run, because:

 - ActiveSetup command from product 123 will be executed, which adds openvpn-gui to Windows\CurrentVersion\Run

 - ActiveSetup command from product 456 will be executed, which removes openvpn-gui from Windows\CurrentVersion\Run

To fix that, we use ">OpenVPN_UserSetup" as ActiveSetup command name, which is guaranteed to be run after all GUID name, similar to NSIS implementation https://github.com/OpenVPN/openvpn-build/commit/68d9ebf52de6be160f813297da62d43b5239ff0b

Also to not to overwrite ActiveSetup command when removing old version during upgrade, make ActiveSetup action no-op in this case.

This fixes https://github.com/OpenVPN/openvpn-build/issues/214

### DIFF
--- a/windows-msi/script/ActiveSetupCA.js
+++ b/windows-msi/script/ActiveSetupCA.js
@@ -24,9 +24,19 @@
  */
 function EvaluateActiveSetup()
 {
-    var
-        productCode = Session.Property("ProductCode"),
-        version;
+    if (Session.EvaluateCondition("UPGRADINGPRODUCTCODE") == 1) {
+        // this means upgrade removes an app - no need to do anything
+        return 1;
+    }
+
+    // we use SimpleProductName (like OpenVPN) which is persistent across upgrades as productCode.
+    // otherwise on upgrade we get a different product code and "uninstall" entry for the old version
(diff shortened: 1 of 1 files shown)


lstipakov:
Also ping @selvanair

selvanair:
Unfortunately I'm cannot to test this right away[*] if you need this for rc2. But the reasoning and the change looks good to me. 
The last time I looked into active setup was ages ago when I implemented it for NSIS -- not sure what the std practice for MSI installers is.

[*] My longish new year break that I squandered on hacking OpenVPN is over...

flichtenheld (approved):

PR #326, opened 2023-01-19
Install ovpn-dco from MSM

lstipakov:
ovpn-dco installation logic has been moved into MSM so that it can be reused by Connect and both openvpn-gui and Connect could co-exist and use the same driver.

PR #327, opened 2023-01-23
Install ovpn-dco from MSM

lstipakov:
ovpn-dco installation logic has been moved into MSM so that it can be reused by Connect and both openvpn-gui and Connect could co-exist and use the same driver.


flichtenheld:
JFTR this is basically #326 but for release/2.6 instead of master.

flichtenheld (approved):

PR #328, opened 2023-01-26
GHA: download strawberry perl from our own mirror

flichtenheld:
Official download location is very unreliable.


lstipakov (approved):
Looks good.

`Using cached strawberry-perl-5.32.1.1-32bit.zip.`

PR #335, opened 2023-02-10
OpenSSL 1.1.1t

flichtenheld:


lstipakov (approved):
Installed on arm64. Checked that openssl version is 1.1.1t and managed to connect.

PR #342, opened 2023-03-16
Work-around GHA build problems

flichtenheld:


lstipakov (approved):
I think we both are not super happy with this change, since it breaks when we would need to bump dependencies/vcpkg. But we do not have a proper solution right now, so let's use this for a while.

PR #349, opened 2023-04-04
msi: configure OpenVPN Service also on reinstall

lstipakov:
We configure openvpn service (set startup mode and run, if needed) only on installation or upgrade. This is achieved by running configuration custom action with "NOT Installed" condition.

However the product may be reinstalled (even though we disable it in UI), for example when pushed via AD Group Policy or with command line

    msiexec /fomusv "OpenVPN-2.6.2-I001-amd64.msi" /L*V reinstall.log

In this case OpenVPN service status is set to "disabled", since our configuration custom action won't be run.

Fix by changing action condition to NOT REMOVE, which covers both install/upgrade and reinstall cases.

Fixes https://github.com/OpenVPN/openvpn-build/issues/348

Change-Id: I92974dea4749c415e25ab0f804050e3e6d309ffc

### DIFF
--- a/windows-msi/msi.wxs
+++ b/windows-msi/msi.wxs
@@ -100,7 +100,7 @@
         <InstallExecuteSequence>
             <Custom Action="FindSystemInfo" After="FindRelatedProducts"/>
             <Custom Action="CheckOpenVPNServiceStatus" After="ProcessComponents"/>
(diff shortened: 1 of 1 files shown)


selvanair:
LGTM. I do not understand MSI, but verified that reinstalling with `/fomusv` does disable the service which we do not want. This patch fixes that.

selvanair (approved):

PR #351, opened 2023-04-05
Really fix reinstall

lstipakov:


lstipakov:
Again ping @selvanair

windows-msi/msi.wxs
```diff
@@ -100,7 +100,7 @@
         <InstallExecuteSequence>
             <Custom Action="FindSystemInfo" After="FindRelatedProducts"/>
             <Custom Action="CheckOpenVPNServiceStatus" After="ProcessComponents"/>
-            <Custom Action="ConfigureOpenVPNService" After="StartServices">NOT REMOVE</Custom>
+            <Custom Action="ConfigureOpenVPNService" After="StartServices">NOT Installed OR REINSTALL</Custom>
```
flichtenheld, line 103: Hmm, this seems to be all very dark magic. By reading the documentation I think I have confirmed that "NOT REMOVE" is wrong for the reasons you stated. I'm not sure whether REINSTALL is correct, though. To me it sounds maybe like what you want to check is https://learn.microsoft.com/en-us/windows/win32/msi/upgradingproductcode ?

flichtenheld (changes requested):
Please also rebase this PR on release/2.6 now that the DCO update is already merged

windows-msi/msi.wxs
lstipakov: On upgrade we use `NOT Installed`, like before. The problem we have is with repair, in which case `REINSTALL` is a correct one according to [this](https://stackoverflow.com/a/17608049). I also tested both cases (unlike previous PR, sigh).

lstipakov:
Rebased.

windows-msi/msi.wxs
flichtenheld: okay

flichtenheld (approved):

PR #357, opened 2023-04-18
debian-sbuild: Release community DCO packages and add GHA tests

flichtenheld:
High level change list:

- Do not release any packages for stretch and bionic anymore (This avoids problems with having distros that do not support DCO)
- Release DCO packages for all distros so that users can easily test latest openvpn2 with latest DCO
- Add GHA builds of debian-sbuild, so we can test changes outside of release builds (this requires changes in release/ as well)
- During Release builds verify signature of Release tarballs

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -227,7 +227,7 @@ jobs:
         working-directory: debian-sbuild
         run: |
           [ ! -f "$CHROOT_CONF" ] || ( tar -xvf "$CHROOT_CONF" -C chroots; sudo install -m644 chroots/chroot.d/* /etc/schroot/chroot.d/ )
-          sudo chown root:root chroots/*.tar.gz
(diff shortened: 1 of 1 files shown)


lstipakov:
I don't feel qualified to review debian packaging, but I noticed that this step took 46min. This noticeably increases the build time. Is this something we really need to do in GHA? Do we expect people to try out packages from CI?

uddr (approved):
LGTM

flichtenheld:
> I don't feel qualified to review debian packaging, but I noticed that this step took 46min. This noticeably increases the build time. Is this something we really need to do in GHA? Do we expect people to try out packages from CI?

I do save the artifacts, so it is something we could do. But for now this is more for my benefit. Note that this will become much faster on rerun since I cache the generated chroots.

lstipakov (approved):

PR #360, opened 2023-04-20
windows-msi: don't depend on core.autocrlf=false

### LINKED ISSUE #358: Windows sample configurations files with wrong EOL and blank lines
Hi,

On Windows, OpenVPN sample config files (server.ovpn and client.ovpn) have problems since few versions :

1) Macintosh (CR) EOL instead of Windows (CR LF) EOL.
(you can check that with Notepad++ editor, for example)
We must convert to Windows (CR LF), otherwise, OpenVPN don't start.

2) Too many blank lines into thoses files, even into comments

(Bug is there since few version, and still there in OpenVPN 2.6.3)

Thank you

flichtenheld:
The build script assumed that the sample files in
openvpn were checked out with LF line-endings. Since these files have no eol enforced by .gitattributes in openvpn, this depends on core.autocrlf=false, which is not the default on Windows.

So assume that these files will be checked out with CRLF line endings on Windows. The alternative would be to enforce LF line-endings via openvpn's
.gitattributes.

Fixes #358

### DIFF
--- a/windows-msi/build.wsf
+++ b/windows-msi/build.wsf
@@ -207,15 +207,15 @@ clean	Cleans intermediate and output files</example>
                             [
                                 BuildPath(depsPath, "openvpn", "COPYING"),
                                 BuildPath(depsPath, "openvpn", "COPYRIGHT.GPL")
-                            ], "utf-8", adLF,
+                            ], "utf-8", adCRLF,
                             []));
                         b.pushRule(new ConvertTextBuildRule(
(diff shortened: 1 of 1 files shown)


flichtenheld:
@lstipakov Do you have an opinion on whether we should fix it in this way or via .gitattributes in openvpn itself?

lstipakov:
I don't have a strong opinion, but since we already have facilities in installer code (which use modern ActiveX technology) I think it is fine to use them.

lstipakov (approved):

PR #372, opened 2023-05-29
Bump Easy-RSA to version 3.1.4

TinCanTech:
Omit Easy-RSA version 3.1.3:
Easy-RSA version 3.1.3 is unable to build a CA using Windows, due to use of file-descriptors, which OpenSSL built for Windows does not support. Easy-RSA ChangeLog has further details.


TinCanTech:
This version is also not ready.

Follow-up: https://github.com/OpenVPN/easy-rsa/issues/962

PR #448, opened 2023-09-04
msi: adapt to openvpn-gui binary dir changes

lstipakov:
Change-Id: If5820b6a1501e7165a063e61c679d4dcf266b519


flichtenheld (changes requested):
Needs changes in build script as well. See GHA build failure

lstipakov:
But this is expected because openvpn-gui submodule is not updated, or am I missing something? I left submodule update out because renovate takes care of it.

flichtenheld:
> But this is expected because openvpn-gui submodule is not updated, or am I missing something? I left submodule update out because renovate takes care of it.

The build failure is with the new openvpn-gui, because we still have `git submodule update --init --remote` in our GHA, i.e. we always build the latest code in GHA. I want to remove that, but didn't get around to it, yet. Either way, a PR needs to build successfully.

flichtenheld (approved):

PR #573, opened 2024-02-12
debian-sbuild: drop support for Ubuntu 23.04 (Lunar)

flichtenheld:
Not supported anymore.

Also remove remaining files for Ubuntu 22.10 (kinetic). We dropped the build already but didn't remove the files.


flichtenheld:
Forgot to do this in time for 2.6.9. So this PR to make sure we do it for 2.6.10.

PR #585, opened 2024-03-06
debian-sbuild: disable systemd.patch

flichtenheld:
Now applied in master.


cron2:
Looks straightforward and reasonable "just remove application of that patch".  Github does not consider anyone but @uddr to be a valid reviewer, so I can't click that button :-)

PR #625, opened 2024-04-18
Prepare release of 2.6.10-I002

uddr:


flichtenheld:
This looks wrong from a history point. This is branched from tip of release/2.6 instead 2.6.10-I001. So it contains src/openvpn that was probably not used for build. Not sure whether we want to clean this up, because this then would mean that we tag something that was not the actual commit used for build. Let's discuss. But definitely just want to merge this.

flichtenheld (changes requested):
see comment

uddr:
@flichtenheld this is a good question

So I took `OpenVPN-2.6.10-I001` tag as a base, added there 'update dependency openvpn/ovpn-dco-win to v1.1.1'
And based on this I created the release and tagged it as `OpenVPN-2.6.10-I002`
Then created from `OpenVPN-2.6.10-I002` tag this `yuriy/msi-I002-build` branch and rebased `release/2.6` into

What is the best way to have it done properly?
Maybe include in this commit all submodules and other changes we had during the release? Get it equal to `OpenVPN-2.6.10-I001` tag? 
And then in the separate commit cover the latest updates we have in `release/2.6`?

uddr:
got it, thanks!
updated

flichtenheld (approved):

PR #628, opened 2024-04-25
Ubuntu noble

flichtenheld:


uddr:
it seems DCO for noble is not linked

flichtenheld:
> it seems DCO for noble is not linked

Yeah, DCO package was completely missing. Added now

PR #631, opened 2024-04-29
Merge 2.6

flichtenheld:
* 2.6.10-I002
* WiX 3.14.1
* MSI_ONLY_BUILD
* Noble packages


flichtenheld:
@uddr Your changes to release/ scripts break GHA build in master (Since there is no such tag as OpenVPN-2.7_git-I001). Can you take a look and advise how we should fix this?
We could either
* change it so that this build doesn't try to create the tag or
* configure the GHA environment in a way that allows the tagging to succeed.

uddr:
@flichtenheld yeah, didn't expect that)
so we usually have this case when we do MSI builds only
I added additional condition for that which should fix the GHA builds
https://github.com/OpenVPN/openvpn-build/pull/632

uddr (approved):

PR #632, opened 2024-04-30
MSI release: tag openvpn-gui for MSI build only

uddr:


release/version-and-tags.sh
```diff
@@ -74,9 +74,12 @@ if ! git diff --exit-code; then
 else
     # If GUI_CURRENT_FULL_VERSION is the same we still want to try to tag OpenVPN build version
     # assuming it might be different and not tagged yet
-    git remote update
-    if ! git tag |grep -q "OpenVPN-$BUILD_VERSION" ; then
-        git tag -a "OpenVPN-$BUILD_VERSION" -m "OpenVPN-$BUILD_VERSION"
+    # This case is usually happens when we need MSI build only
+    if [[ "$MSI_BUILD_ONLY" == "YES" ]]; then
```
flichtenheld, line 78: No, let's not do this. Let's also handle the unusual case correctly. We definitely had this several times.
Let's fix GHA instead by setting environment variables or writing a config file.

flichtenheld (changes requested):

release/version-and-tags.sh
uddr: makes sense
created PR for GHA https://github.com/OpenVPN/openvpn-build/pull/633/files
thanks

PR #633, opened 2024-05-01
GHA: git: configure GHA user name

uddr:
We need to have ability to tag the builds.
Updated release flow tries to tag openvpn-gui
if this tag does not exist.
Before that was done manually.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -195,6 +195,9 @@ jobs:
         run: |
           ln -s vars.example vars
           mkdir ../output
+          # we need to have ability to git tag the versions
+          # we do not push that tags
+          git config --global user.name "gha_user"
           ./version-and-tags.sh
           ./create-release-files.sh
 


.github/workflows/build.yaml
```diff
@@ -195,6 +195,9 @@ jobs:
         run: |
           ln -s vars.example vars
           mkdir ../output
+          # we need to have ability to git tag the versions
+          # we do not push that tags
+          git config --global user.name "gha_user"
```
flichtenheld, line 200: This is not enough. Email also needs to be configured. Still fails with `fatal: unable to auto-detect email address (got 'runner@fv-az1149-423.(none)')`
You need to merge your changes on top of my merge branch and then test the build to see whether it works. Because this only affects master builds and not release/2.6 builds.

flichtenheld (changes requested):

flichtenheld:
Taking this over since @uddr is on vacation.

Issue #671, opened 2024-06-22
Updating OpenVPN on Windows fails when the OS language is other than English


kevloral:
Beginning in 2.6.9, when you try to update OpenVPN on Windows and a non standard installation folder has been used, the process always fails with this message:

![image](https://github.com/OpenVPN/openvpn-build/assets/3647248/3a607511-be92-4621-8fa6-d19f06f7b8c5)

I assume the problem lies in the new script that was introduced in that version intended to reset the permissions of the installation folder (I guess it may not support using non standard destinations).

If the installer is executed with the /L option, you get this in the resulting log file:

```
[MSI (s) (3C:B4) [00:02:53:985]: Executing op: ActionStart(Name=SetACL,,)
Action 0:02:53: SetACL. 
MSI (s) (3C:B4) [00:02:53:987]: Executing op: CustomActionSchedule(Action=SetACL,ActionType=7173,Source=/*
 *  openvpn-build â€” OpenVPN packaging
 *
 *  Copyright (C) 2024 Lev Stipakov <lev@openvpn.net>
 *
 *  This program is free software; you can redistribute it and/or modify
 *  [..]
 */

function InProgramFiles(installPath) {
    var shell = new ActiveXObject("WScript.Shell");
    var fso = new ActiveXObject("Scripting.FileSystemObject");

    var programFilesPath = shell.ExpandEnvironm
Error 1720. There is a problem with this Windows Installer package. A script required for this install to complete could not be run. Contact your support personnel or package vendor.  Custom action SetACL script error -2147467259, Msi API Error: Property,Name Line 176, Column 9,  
MSI (s) (3C:54) [00:02:55:944]: Product: OpenVPN 2.6.9-I001 amd64 -- Error 1720. There is a problem with this Windows Installer package. A script required for this install to complete could not be run. Contact your support personnel or package vendor.  Custom action SetACL script error -2147467259, Msi API Error: Property,Name Line 176, Column 9,  ]
```

The problem also affects versions 2.6.10 and 2.6.11

kevloral:
I dug into it a little bit and it is not that the installation fails when the destination is not the default one, but when the OS is not in English. 

Given that the error message in the log file is quite useless (_Property,Name Line 176, Column 9_), perhaps because there is a problem in the way the _SetACL_ function tries to return an error message, I fired up a process monitor to find out the exact _icacls_ parameters the installer was trying to use. In my case, this is the full command:

[...]

PR #674, opened 2024-06-25
msi: use SIDs instead of built-in group names

lstipakov:
Group names are localized in some localizations, so we have to use SIDs.

Fixes https://github.com/OpenVPN/openvpn-build/issues/671.

### DIFF
--- a/windows-msi/script/ACL.js
+++ b/windows-msi/script/ACL.js
@@ -171,7 +171,8 @@ function SetACL() {
     targetDir = TrimTrailingSlash(targetDir);
 
     try {
-        RunIcacls('\"' + targetDir + '\" /inheritance:r /grant "Administrators:(OI)(CI)F" /grant "System:(OI)(CI)F" /grant "Users:(OI)(CI)RX"');
+        // Administrators, SYSTEM, Users
+        RunIcacls('\"' + targetDir + '\" /inheritance:r /grant "*S-1-5-32-544:(OI)(CI)F" /grant "*S-1-5-18:(OI)(CI)F" /grant "*S-1-5-32-545:(OI)(CI)RX"');
     } catch (e) {
         Session.Property("CUSTOMACTIONERROR") = e.message;
         return 1603; // Indicates a fatal error during installation.


cron2:
code change is in line with the documentation under https://learn.microsoft.com/en-us/windows/win32/secauthz/well-known-sids - so, looks good to me.  Would be good to have a "now it works" confirmation from #671

windows-msi/script/ACL.js
```diff
@@ -171,7 +171,8 @@ function SetACL() {
     targetDir = TrimTrailingSlash(targetDir);
 
     try {
-        RunIcacls('\"' + targetDir + '\" /inheritance:r /grant "Administrators:(OI)(CI)F" /grant "System:(OI)(CI)F" /grant "Users:(OI)(CI)RX"');
+        // Administrators, SYSTEM, Users
+        RunIcacls('\"' + targetDir + '\" /inheritance:r /grant "*S-1-5-32-544:(OI)(CI)F" /grant "*S-1-5-18:(OI)(CI)F" /grant "*S-1-5-32-545:(OI)(CI)RX"');
```
selvanair, line 175: While this looks good, we are only removing inherited permissions and setting new explicit ones. Any explicit permission that was already present will remain unchanged. Is this what we want? Or should we be resetting acls first and then setting this? 

For example: with 2.6.8 installed in to a folder with `foo\selva:(OCI)(CI)F` in addition to inherited acls,
on upgrading to 2.6.11 the above explicit permission will be retained and new explicit one's for SYSTEM, Users and Administrators added.

lstipakov: This specific commit doesn't alter the logic. It just has slipped from my mind that build-in groups might have localized names (and it is not the case for my Finnish localization by some reasons).

One can argue that if the user has explicitly assigned a permission to the installation directory he knows what he is doing. On the other hand indeed it could weaken the security.

Note that the impact of this code is somewhat limited, since starting from 2.6.9 users are not able to customize the installation directory.

selvanair: Yes, I was asking something I should have asked in response to the original commit in January -- did not occur to me that time :)

As for the fix here to use SIDs to avoid the issue of localization of group names, looks good to me though I do not have a localized system to test this on.

lstipakov:
Got a confirmation from a colleague with Brazilian Portuguese localization that this indeed fixes the problem.

PR #675, opened 2024-06-26
msi: use SIDs instead of built-in group names

lstipakov:
Group names are localized in some localizations, so we have to use SIDs.

Fixes https://github.com/OpenVPN/openvpn-build/issues/671.


cron2 (approved):
Same change as already approved for master -> ACK.

PR #828, opened 2025-02-21
renovate: update ovpn-dco-win only within 1.x versions

lstipakov:


flichtenheld:
https://docs.renovatebot.com/configuration-options/ says

> Renovate always uses the config from the repository's default branch, even if that configuration specifies multiple baseBranches. Renovate does not read/override the config from within each base branch if present.

So this will not work, I'm afraid.

PR #832, opened 2025-02-21
renovate: ensure correct dco-win versioning in master and release/2.6…

lstipakov:
… branches

It seems that renovate uses config from default branch, so we have to ensure versioning for release branch here.


flichtenheld (approved):
LGTM

PR #894, opened 2025-04-30
Switch from ovpn-dco to ovpn-backports

flichtenheld:


flichtenheld:
Should be merged after https://gerrit.openvpn.net/c/openvpn/+/941

windows-msi/version.m4
```diff
@@ -42,15 +42,15 @@ define([PRODUCT_NAME],      [OpenVPN])
 define([PRODUCT_PUBLISHER], [OpenVPN, Inc.])
 
 dnl The package version as displayed by UI and used in filenames (no spaces, please).
-define([PACKAGE_VERSION], [2.7git])
+define([PACKAGE_VERSION], [2.7_git-I001])
```
ordex, line 45: is this and (and below) an unrelated change?

uddr (approved):

windows-msi/version.m4
lstipakov: Indeed.

flichtenheld: Yes. When testing the build locally the tools updated this. I left it in although it makes no real difference, since it makes sense to me having the PACKAGE_VERSION closer to what we actually use.

cron2:
> Should be merged after https://gerrit.openvpn.net/c/openvpn/+/941

941 is in!

PR #903, opened 2025-05-12
msi: fix PRODUCT_VERSION in version.m4

lstipakov:
Commit

  644b72ca ("Switch from ovpn-dco to ovpn-backports")

by some reasons has changed PRODUCT_VERSION from 2.7.0 to 2.7_git01, which broke Wix installer build. This hasn't been catched by GHA since in GHA this value is overwritten.

Fix by reverting back to original, working value.


flichtenheld (changes requested):
Please also update `release/vars.example` so that this can't happen again.
Also while we're on this maybe also update GHA to use 2.7 version number.

flichtenheld (approved):
As discussed on IRC, apply simple revert for now and worry about how to fix vars.example later

uddr (approved):

PR #929, opened 2025-06-03
windows-msi: Remove support building .exe installer

flichtenheld:
This hasn't been used in years and probably bitrotted badly. Let's just remove it to not pretend this is supported.


windows-msi/build.wsf
```diff
@@ -329,28 +329,6 @@ clean	Cleans intermediate and output files</example>
                     break;
             }
 
-            // Add rules for 7-Zip SFX creation.
```
mattock, line 332: There's a *case "exe"* just above this (lines 326-329). Should it be removed also?

flichtenheld: Removed more stuff :)

PR #944, opened 2025-06-18
msi: fix service log directory permissions

lstipakov:
Commit

    42bf901 ("msi: change log directory for persistent connections")

changed log directory for persistent connections, but haven't explicitly added permissions to virtual service account running those connections.

This adds missing permissions.

Github: #942

Reported-by: Jernej Simončič


selvanair:
> Would all users have read access to the log? We need that for "view log" to work from the GUI when attached to the persistent connection.

Replying to self. It seems the "alpha1" installer appears to works for me even without this patch and log files (new as well as old copied over ones) are owned by the new service account (!!) and all users have read access.

PR #1125, opened 2025-11-02
Register an event source for logging from the GUI

selvanair:
Goes with OpenVPN/openvpn-gui#776


selvanair:
Do not merge this yet. If we decide to reuse openvpnservmsg.dll for this, that would need a different patch.

flichtenheld:
> Rebased to 2.7. How do I change the target of the PR now?

Via the "Edit" button in the upper right corner.

PR #1152, opened 2025-11-28
windows-msi: make Windows 10 minimum supported version

lstipakov:


cron2:
I have no idea what this all does, but the change looks logical to me (assuming that windows versions are somewhat logical in this context).

selvanair:
Is `VersionNT=1000`,  the correct value for Windows 10? There are conflicting info out there: this page says 603 (same as Windows 8.1?) : https://learn.microsoft.com/en-us/troubleshoot/windows-client/application-management/versionnt-value-for-windows-10-server

flichtenheld (approved):

PR #1329, opened 2026-04-21
debian-sbuild: New Ubuntu 26.04 resolute; cleanup old distros

uddr:


cron2 (approved):
the changelog bit is a bit confusing for me ("why do we have 2.6.14-ubuntu things here, when we're working on a 2.6.20-openvpn release?") but this is not "new", so maybe needs to be clarified separately

